PR #41, opened 2017-11-05
(22) Restricted Line Feed Characters in PskMessage Headers

controlsurface:
Updated PskMessageBinaryCode to reject encoding or decoding PskMessages with headers containing line feeds.


adrianhopebailie:
@controlsurface this looks great thanks.

Please can you amend the commit to include your signature and then push it again. That will get the DCO check to pass.

You should be able to simply do a `git commit --amend -s` then `git push --force`

sappenin:
@adrianhopebailie Seems like PSKv1 is dead now - can we close this?

sappenin:
PSKv1 is deprecated.

PR #48, opened 2017-11-14
Fixes #45, Fixes #46, Fixes #47

### LINKED ISSUE #45: InterledgerAddress#getPrefix Fails on a root prefix
We need to tweak InterledgerAddress#getPrefix to work properly on something like `g.` or any other root prefix. In this case, the method should probably return `Optional#empty`.

### LINKED ISSUE #46: Add InterledgerAddress#getParentPrefix
`InterledgerAddress.of("g.b.").getPrefix()` should return `InterledgerAddress.of("g.")` but instead returns `InterledgerAddress.of("g.b.").getPrefix()`.

### LINKED ISSUE #47: Implement InterledgerAddress#hasPrefix
It would be convenient to have a method `InterledgerAddress#hasParentPrefix` which returns true if an address has a parent prefix. The address `g.` would return false.

Alternatively, we could implement `InterledgerAddress#isRootPrefix`.

sappenin:
* Fixes #45.
* Fixes #46.
* Fixes #47.
* Fixes checkstyle configuration, and actually runs checkstyle during the build.


dev-ops/src/main/resources/org/interledger/dev-ops/checkstyle.xml
```diff
@@ -59,7 +59,7 @@
         </module>
         <module name="NeedBraces"/>
         <module name="LeftCurly">
-            <property name="maxLineLength" value="100"/>
+            <!--<property name="maxLineLength" value="100"/>-->
```
sappenin: Inclusion of this property in the checkstyle file breaks the Intellij Checkstyle Plugin. However, if people don't like this, I'm open to restoring the line.

sappenin:
Closing in-favor of #49.

PR #55, opened 2017-11-21
Use immutable

adrianhopebailie:
Using Immutable for InterledgerAddress
Fixed a lot of Checkstyle violations

### DIFF
--- a/crypto-conditions/src/main/java/org/interledger/cryptoconditions/CryptoConditionReader.java
+++ b/crypto-conditions/src/main/java/org/interledger/cryptoconditions/CryptoConditionReader.java
@@ -1,12 +1,13 @@
 package org.interledger.cryptoconditions;
 
-import net.i2p.crypto.eddsa.EdDSAPublicKey;
-import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
-import net.i2p.crypto.eddsa.spec.EdDSAPublicKeySpec;
 import org.interledger.cryptoconditions.der.DerEncodingException;
 import org.interledger.cryptoconditions.der.DerInputStream;
 import org.interledger.cryptoconditions.der.DerTag;
 import org.interledger.cryptoconditions.utils.UnsignedBigInteger;
+
+import net.i2p.crypto.eddsa.EdDSAPublicKey;
+import net.i2p.crypto.eddsa.spec.EdDSANamedCurveTable;
+import net.i2p.crypto.eddsa.spec.EdDSAPublicKeySpec;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
@@ -34,9 +35,7 @@ public class CryptoConditionReader {
    * Reads a DER encoded condition from the buffer.
    *
    * @param buffer contains the raw DER encoded condition.
-   *
    * @return The condition read from the buffer.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    */
   public static Condition readCondition(byte[] buffer) throws DerEncodingException {
@@ -49,9 +48,7 @@ public class CryptoConditionReader {
    * @param buffer contains the raw DER encoded condition.
    * @param offset the position within the buffer to begin reading the condition.
    * @param length the number of bytes to read.
-   *
    * @return The condition read from the buffer.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    */
   public static Condition readCondition(byte[] buffer, int offset, int length)
@@ -77,9 +74,7 @@ public class CryptoConditionReader {
    * Reads a DER encoded condition from the input stream.
    *
    * @param in The input stream containing the DER encoded condition.
-   *
    * @return The condition read from the stream.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    * @throws IOException          if any I/O operation fails for any reason.
    */
@@ -93,9 +88,7 @@ public class CryptoConditionReader {
    *
    * @param in        The input stream containing the DER encoded condition.
    * @param bytesRead will be updated with the number of bytes read from the stream.
-   *
    * @return The condition read from the stream.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    * @throws IOException          if any I/O operation fails for any reason.
    */
@@ -139,9 +132,7 @@ public class CryptoConditionReader {
    * Reads a DER encoded fulfillment from the buffer.
    *
    * @param buffer The buffer holding the DER encoded fulfillment
-   *
    * @return The fulfillment read from the buffer.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    */
   public static Fulfillment readFulfillment(byte[] buffer) throws DerEncodingException {
@@ -154,9 +145,7 @@ public class CryptoConditionReader {
    * @param buffer The buffer holding the DER encoded fulfillment
    * @param offset the position within the buffer to begin reading the fulfilment.
    * @param length the number of bytes to read.
-   *
    * @return The fulfillment read from the buffer.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    */
   public static Fulfillment readFulfillment(byte[] buffer, int offset, int length)
@@ -182,9 +171,7 @@ public class CryptoConditionReader {
    * Reads a DER encoded fulfillment from the input stream.
    *
    * @param in The input stream containing the DER encoded fulfillment.
-   *
    * @return The fulfillment read from the stream.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    * @throws IOException          if any I/O operation fails for any reason.
    */
@@ -198,9 +185,7 @@ public class CryptoConditionReader {
    *
    * @param in        The input stream containing the DER encoded fulfillment.
    * @param bytesRead will be updated with the number of bytes read from the stream.
-   *
    * @return The fulfillment read from the stream.
-   *
    * @throws DerEncodingException when DER encoding fails for any reason.
    * @throws IOException          if any I/O operation fails for any reason.
    */
--- a/crypto-conditions/src/main/java/org/interledger/cryptoconditions/RsaSha256Fulfillment.java
+++ b/crypto-conditions/src/main/java/org/interledger/cryptoconditions/RsaSha256Fulfillment.java
@@ -59,8 +59,7 @@ public class RsaSha256Fulfillment extends FulfillmentBase<RsaSha256Condition>
   /**
    * Returns a copy of the signature used in this fulfillment.
    *
-   * @return A {@link byte[]} containing the signature for this fulfillment.
-   *
+   * @return A byte array containing the signature for this fulfillment.
    * @deprecated Java 8 does not have the concept of an immutable byte array, so this method allows
    *     external callers to accidentally or intentionally mute the prefix. As such, this method may
    *     be removed in a future version. Prefer {@link #getSignatureBase64Url()} instead.
--- a/ilp-core/src/main/java/org/interledger/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/InterledgerAddress.java
@@ -1,5 +1,7 @@
 package org.interledger;
 
+import org.interledger.annotations.Immutable;
+
 import org.immutables.value.Value;
 
 import java.util.List;
@@ -30,7 +32,6 @@ import java.util.regex.Pattern;
  *
  * @see "https://github.com/interledger/rfcs/tree/master/0015-ilp-addresses"
  */
-@Value.Immutable
 public interface InterledgerAddress {
 
   String REGEX = "(?=^.{1,1023}$)"
@@ -46,9 +47,7 @@ public interface InterledgerAddress {
    * @return an {@link InterledgerAddress} instance.
    */
   static InterledgerAddress of(final String value) {
-    return builder()
-        .value(value)
-        .build();
+    return new InterledgerAddressBuilder().value(value).build();
   }
 
   /**
@@ -137,10 +136,10 @@ public interface InterledgerAddress {
   /**
    * Get the default builder.
    *
-   * @return a {@link ImmutableInterledgerAddress.Builder} instance.
+   * @return a {@link InterledgerAddressBuilder} instance.
    */
-  static ImmutableInterledgerAddress.Builder builder() {
-    return ImmutableInterledgerAddress.builder();
+  static InterledgerAddressBuilder builder() {
+    return new InterledgerAddressBuilder();
   }
 
   /**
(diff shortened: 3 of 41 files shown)


ilp-core/src/main/java/org/interledger/InterledgerPacket.java
```diff
@@ -39,7 +38,7 @@
      * A handler for allowing callers to specify logic based upon an unknown result type. This class
      * can be used in the following manner:
      *
-     * <pre>
+     * <p><pre>
```
sappenin: I get a checkstyle build failure with this -- does this build properly on your local machine?

ilp-core/src/main/java/org/interledger/InterledgerPacket.java
```diff
@@ -80,7 +77,6 @@
        * Handle an instance of {@link QuoteLiquidityResponse}.
        *
        * @param quoteLiquidityResponse An instance of {@link QuoteLiquidityResponse}.
-       *
```
sappenin: My code-styler has these newlines by default - are sure this is correct?

crypto-conditions/src/main/java/org/interledger/cryptoconditions/RsaSha256Fulfillment.java
```diff
@@ -59,8 +59,7 @@ public RSAPublicKey getPublicKey() {
   /**
    * Returns a copy of the signature used in this fulfillment.
    *
-   * @return A {@link byte[]} containing the signature for this fulfillment.
-   *
```
sappenin: See my comment below - I think the newline should be kept between `@return` and `@deprecated`. Below is a better example.

crypto-conditions/src/main/java/org/interledger/cryptoconditions/CryptoConditionReader.java
```diff
@@ -34,9 +35,7 @@
    * Reads a DER encoded condition from the buffer.
    *
    * @param buffer contains the raw DER encoded condition.
-   *
```
sappenin: Don't think these newlines are correct. See my comment below in `InterledgerPacket`.

ilp-core/src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -1,5 +1,7 @@
 package org.interledger;
 
+import org.immutables.value.Value;
```
sappenin: First of all, this all looks great - nice work! (this has actually been on my todo list for a while).

One thing I've noticed with Immutables is that if you annotate the interface, then you can't extend that interface with another Immutable. In order to do that, we need to use an `abstract` class. It's a _little_ more code for us, but I think we should do it so that somebody could create an extension of our interfaces, but still be able to use the Immutables library. 

One nice effect this has, however, is it allows us to define the `equals` and `hashcode` methods correctly, so that anybody using the library doesn't have to worry about it.

So, `InterledgerAddress.java` would instead look like this:

```
package org.interledger;

import org.immutables.value.Value;

import java.util.List;
import java.util.Objects;
import java.util.Optional;
import java.util.Set;
import java.util.regex.Pattern;

/**
 * Interledger Protocol (ILP) Addresses identify Ledger accounts (or groups of Ledger accounts) in
 * an ILP network, and provide a way to route a payment to its intended destination.
 *
 * <p>Interledger Addresses can be subdivided into two categories:</p>
 *
 * <p> <b>Destination Addresses</b> are complete addresses that can receive payments. A destination
 * address always maps to one account in a ledger, though it can also provide more specific
 * information, such as an invoice ID or a sub-account. Destination addresses MUST NOT end in a
 * period (.) character. </p>
 *
 * <p> <b>Address Prefixes</b> are incomplete addresses representing a grouping of destination
 * addresses. Many depths of grouping are possible, for example: groups of accounts or sub-accounts;
 * an individual ledger or sub-ledger; or entire neighborhoods of ledgers. Address prefixes MUST end
 * in a period (.) character. </p>
 *
 * <p> The formal specification for an Interledger Addresses is defined in <a
 * href="https://github.com/interledger/rfcs/tree/master/0015-ilp-addresses">Interledger RFC
 * #15</a>. </p>
 *
 * @see "https://github.com/interledger/rfcs/tree/master/0015-ilp-addresses"
 */

      + "^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$";

  Pattern PATTERN = Pattern.compile(REGEX);

  /**
   * <p>Constructor to allow quick construction from a String representation of an ILP address.</p>
   *
   * @param value String representation of an Interledger Address
   *
   * @return an {@link InterledgerAddress} instance.
   */
  static InterledgerAddress of(final String value) {
    return builder()
        .value(value)
        .build();
  }

  /**
   * <p>Helper method to determine if an Interledger Address conforms to the specifications outlined
   * in Interledger RFC #15.</p>
   *
   * @param value A {@link String} representing a potential Interledger Address value.
   *
   * @return {@code true} if the supplied {@code value} conforms to the requirements of RFC 15;
   *     {@code false} otherwise.
   *
   * @see "https://github.com/interledger/rfcs/tree/master/0015-ilp-addresses"
   */
  static boolean isValid(final String value) {
    Objects.requireNonNull(value);
    return PATTERN.matcher(value)
        .matches();
  }

  /**
   * Checks and requires that the specified {@code addressPrefix} is an address prefix per {@link
   * InterledgerAddress#isLedgerPrefix()}.
   *
   * <p>This method is designed primarily for doing parameter validation in methods and
   * constructors, as demonstrated below:</p> <blockquote>
   * <pre>
   * public Foo(InterledgerAddress bar) {
   *     this.ledgerPrefix = InterledgerAddress.requireAddressPrefix(bar);
   * }
   * </pre>
   * </blockquote>
   *
   * @param addressPrefix A {@link InterledgerAddress} to check.
   *
   * @return {@code ledgerPrefix} if its value ends with a dot (.).
   *
   * @throws IllegalArgumentException if the supplied Interledger address is not a ledger-prefix.
   */
  static InterledgerAddress requireAddressPrefix(final InterledgerAddress addressPrefix) {
    Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
    if (!addressPrefix.isLedgerPrefix()) {
      throw new IllegalArgumentException(
          String.format("InterledgerAddress '%s' must be an Address Prefix ending with a dot (.)",
              addressPrefix.getValue()
          )
      );
    } else {
      return addressPrefix;
    }
  }

  /**
   * Checks and requires that the specified {@code addressPrefix} is not an address prefix per
   * {@link InterledgerAddress#isLedgerPrefix()}.
   *
   *
   * <p>This method is designed primarily for doing parameter validation in methods and
   * constructors, as demonstrated below:</p> <blockquote>
   * <pre>
   * public Foo(InterledgerAddress bar) {
   *     this.nonLedgerPrefix = InterledgerAddress.requireNotAddressPrefix(bar);
   * }
   * </pre>
   * </blockquote>
   *
   * @param addressPrefix A {@link InterledgerAddress} to check.
   *
   * @return {@code addressPrefix} if its value ends with a dot (.).
   *
   * @throws IllegalArgumentException if the supplied Interledger address is not a ledger-prefix.
   */
  static InterledgerAddress requireNotAddressPrefix(final InterledgerAddress addressPrefix) {
    Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
    if (addressPrefix.isLedgerPrefix()) {
      throw new IllegalArgumentException(
          String
              .format("InterledgerAddress '%s' must NOT be an Address Prefix ending with a dot (.)",
                  addressPrefix.getValue()
              )
      );
    } else {
      return addressPrefix;
    }
  }

  /**
   * Get the default builder.
   *
   * @return a {@link ImmutableInterledgerAddress.Builder} instance.
   */
  static ImmutableInterledgerAddress.Builder builder() {
    return ImmutableInterledgerAddress.builder();
  }

  /**
   * Return this address's value as a non-null {@link String}. For example:
   * <code>us.usd.bank.account</code>
   *
   * @return A {@link String} representation of this Interledger address.
   */
  String getValue();

  /**
   * Tests if this Interledger address represents a ledger prefix.
   *
   * @return True if the address is a ledger prefix, false otherwise.
   */
  default boolean isLedgerPrefix() {
    return getValue().endsWith(".");
  }

  /**
   * Tests if this InterledgerAddress starts with the specified {@code addressSegment}.
   *
   * @param addressSegment An {@link String} prefix to compare against.
   *
   * @return {@code true} if this InterledgerAddress begins with the specified prefix.
   */
  default boolean startsWith(final String addressSegment) {
    Objects.requireNonNull(addressSegment, "addressSegment must not be null!");
    return this.getValue()
        .startsWith(addressSegment);
  }

  /**
   * Tests if this InterledgerAddress starts with the specified {@code interledgerAddress}.
   *
   * @param interledgerAddress An {@link InterledgerAddress} prefix to compare against.
   *
   * @return {@code true} if this InterledgerAddress begins with the specified prefix.
   */
  default boolean startsWith(final InterledgerAddress interledgerAddress) {
    Objects.requireNonNull(interledgerAddress, "interledgerAddress must not be null!");
    return this.startsWith(interledgerAddress.getValue());
  }

  /**
   * <p>Return a new InterledgerAddress by postfixing the supplied {@code segment} to this address.
   * </p>
   *
   * <p>This method can be used to construct both address prefixes and destination addresses. For
   * example, if the value of this address is '<code>us.usd.</code>', then calling this method with
   * an argument of '<code>bob</code>' would result in a new Interledger Address with a value of
   * '<code>us.usd.bob</code>', which is a destination address.</p>
   *
   * <p>Likewise, if the value of this address is '<code>us.usd.pacific.</code>', then calling this
   * method with an argument of '<code>creditunions.</code>' would result in a new Interledger
   * Address with a value of '<code>us.usd.pacific.creditunions.</code>', which is an address
   * prefix.</p>
   *
   * @param addressSegment A {@link String} to be appended to this address as an additional
   *                       segment.
   *
   * @return A new instance representing the original address with a newly specified final segment.
   */
  default InterledgerAddress with(String addressSegment) {
    Objects.requireNonNull(addressSegment, "addressSegment must not be null!");

    final StringBuilder sb = new StringBuilder(this.getValue());
    if (!this.isLedgerPrefix()) {
      sb.append(".");
    }
    sb.append(addressSegment);

    return InterledgerAddress.of(sb.toString());
  }

  /**
   * <p>Return this address's prefix, which is a new {@link InterledgerAddress} containing the
   * characters inside of {@link #getValue()}, up-to and including the last period. If this address
   * is already a prefix, then this instance is instead returned unchanged.</p>
   *
   * <p>For example, calling this method on an address 'g.example.alice' would yield a new address
   * containing 'g.example.'. Conversely, calling this method on an address that is already a
   * prefix, like 'g.example.' would yield the same instance, 'g.example.'.</p>
   *
   * @return A potentially new {@link InterledgerAddress} representing the prefix of this address.
   */
  default InterledgerAddress getPrefix() {
    if (this.isLedgerPrefix()) {
      return this;
    } else {
      return InterledgerAddress.of(getValue().substring(0, this.getValue().lastIndexOf(".") + 1));
    }
  }

  /**
   * <p>Return this address's parent prefix.</p>
   *
   * <p>If this address is a destination address, then this method returns a new {@link
   * InterledgerAddress} containing the characters inside of {@link #getValue()}, up-to and
   * including last period. If this address is instead a prefix, then this instance returns a new
   * {@link InterledgerAddress} containing the characters inside of {@link #getValue()}, up-to and
   * including second-to-last period, unless this address is a root prefix, in which case, {@link
   * Optional#empty()} is returned.</p>
   *
   * <p>For example, calling this method on an address 'g.example.alice' would yield a new address
   * containing 'g.example.'. Likewise, calling this method on an address that is already a prefix,
   * like 'g.example.' would yield 'g.'. Finally, calling this method on a root prefix, like "self."
   * would yield {@link Optional#empty()}.</p>
   *
   * @return An optionally present parent-prefix.
   */
  default Optional<InterledgerAddress> getParentPrefix() {
    // If this address is not a prefix, then just return the prefix. Otherwise, look deeper.
    if (this.isLedgerPrefix()) {
      // If the prefix is a root prefix, return Optional#empty. Otherwise, return the parent prefix.
      if (isRootPrefix()) {
        return Optional.empty();
      } else {
        // Call getParentPrefix with the account portion, which will forward to #getPrefix.
        return InterledgerAddress.of(
            this.getValue().substring(0, this.getValue().lastIndexOf("."))
        ).getParentPrefix();
      }
    } else {
      return Optional.of(this.getPrefix());
    }
  }

  /**
   * <p>Determines if this ILP Address has a parent-prefix.</p>
   *
   * <p>If this address is a destination address, then it has a parent prefix. However, if the
   * address is a prefix, then it only has a parent if it is _not_ a Root Prefix.</p>
   *
   * @return {@code true} if this address is a destination address. Otherwise (if this address is a
   *     prefix address), then return {@code false} if this address is a Root prefix; otherwise,
   *     return {@code true}.
   */
  default boolean hasParentPrefix() {
    // All ILP addresse have a parent prefix, except for Root prefixes.
    return isRootPrefix() == false;
  }

  String ROOT_REGEX = "^(g|private|example|peer|self|test[1-3])[.]$";
  Pattern ROOT_PATTERN = Pattern.compile(ROOT_REGEX);

  /**
   * <p>Determines if this address is a "root" prefix, which per ILP-RFC-15, is one of: "g.",
   * "private.", "example.", "peer.", "self.", "test1.", "test2.", or "test3.". Any other kind of
   * valid ILP address (e.g. "g.1") is not a root prefix.</p>
   *
   * @return {@code true} if this address is a root prefix; {@code false} otherwise.
   */
  default boolean isRootPrefix() {
    // Alternate implementation (Note: A prefix is a root prefix if it has only a single period)
    // final int numPeriods = getValue().length() - getValue().replaceAll("[.!?]+", "").length();
    // return numPeriods == 1;
    return ROOT_PATTERN.matcher(this.getValue()).matches();
  }

  @Value.Immutable
  abstract class AbstractInterledgerAddress implements InterledgerAddress {

    /**
     * <p>Compares the specified object with this <tt>InterledgerAddress</tt> for equality.</p>
     *
     * <p>The <tt>InterledgerAddress</tt> interface is essentially a type-safe wrapper around a
     * String value, so implementations should take care to forward equality decisions to the {@link
     * String#equals(Object)} method on the object returned by {@link #getValue()}.</p>
     *
     * @param obj object to be compared for equality with this collection
     *
     * @return <tt>true</tt> if the specified object is equal to this {@link InterledgerAddress}.
     *
     * @see Object#equals(Object)
     * @see Set#equals(Object)
     * @see List#equals(Object)
     */
    @Override
    public boolean equals(Object obj) {
      if (this == obj) {
        return true;
      }
      if (obj == null || getClass() != obj.getClass()) {
        return false;
      }

      InterledgerAddress foo = (InterledgerAddress) obj;

      return getValue().equals(foo.getValue());
    }

    /**
     * <p>Returns the hash code value for this <tt>InterledgerAddress</tt>.</p>
     *
     * <p>The<tt>InterledgerAddress</tt> interface is essentially a type-safe wrapper around a
     * String value, so implementations should take care to forward hashcode decisions to the {@link
     * String#equals(Object)} method on the object returned by {@link #getValue()}.</p>
     *
     * @return the hash code value for this {@link InterledgerAddress}.
     *
     * @see Object#hashCode()
     * @see Object#equals(Object)
     */
    @Override
    public int hashCode() {
      return this.getValue().hashCode();
    }

    /**
     * Precondition enforcer that ensures the value is a valid Interledger Address.
     *
     * @see "https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md"
     */
    @Value.Check
    void check() {
      if (!InterledgerAddress.isValid(getValue())) {
        throw new IllegalArgumentException(String.format("Invalid characters in address: ['%s']. "
            + "Reference Interledger ILP-RFC-15 for proper format.", getValue()));
      }
    }
  }
}
``` 

To make the above work, you'll need to edit `CodecContext#lookup` to be as follows:

```
/**
   * Helper method to lookup a {@link Codec} for the specified {@code type}.
   *
   * @param <T>  The specific type of {@link Codec} to return.
   * @param type An instance of {@link Class}.
   */
  @SuppressWarnings("unchecked")
  private <T> Codec<T> lookup(final Class<T> type) {
    Objects.requireNonNull(type);

    if (codecs.containsKey(type)) {
      return (Codec<T>) codecs.get(type);
    } else {
      // Check any interfaces...
      return Optional.ofNullable(
          Arrays.stream(type.getInterfaces())
              // Only use codecs that are found...
              .filter(codecs::containsKey)
              // If found, map to the first codec...
              .map(interfaceClass -> (Codec<T>) codecs.get(interfaceClass))
              .findFirst()
              // Otherwise, recurse with the super-class
              .orElseGet(() -> {
                // There were not interfaces, so check the super-classes.
                return (Codec<T>) lookup(type.getSuperclass());
              })
      ).orElseThrow(() -> new CodecException(
          String.format("No codec registered for %s or its super classes!",
              type.getName())));
    }
  }
```

adrianhopebailie:
I have struggled to make this work as you have it here.

I am inclined to agree with @earizon that maybe we should move to using Abstract classes instead of Interfaces for many of these.

All of the classes we have in here are business objects so they are probably better suited to being Abstract classes.  I'm going to update this to try that pattern and let's see what everyone thinks.

adrianhopebailie:
Have not change crypto-conditions but wanted to get feedback on the pattern.
Basically the idea is for any business object:
1. Define the interface
2. Add an internal abstract implementation with definitions for equals/hashcode etc
3. Decorate the abstract implementation with `@Immutable` annotation
4. Add a static method to the interface that returns a new instance of XBuilder

So if you have an interface called `MyThing` this gives developers an interface an base abstract class (`MyThing.AbstractMyThing`) that they can extend and not need to implement hashcode/equals etc and a static `MyThing.builder()` that returns a `MyThingBuilder`

adrianhopebailie:
Ping @sappenin (have made changes based on discussion in the call yesterday)

sappenin:
I fixed the minor nits that I had per this commit (https://github.com/hyperledger/quilt/pull/55/commits/136f6dfa9b9f2fcd07faba55dd272a9ba9639639). Should be good to merge now if you're ok with my small changes.

adrianhopebailie:
@sappenin - Thanks. I had to replace your commit to get past the DCO check.

PR #70, opened 2018-01-10
Fixes #10

### LINKED ISSUE #10: [crypto-conditions] Add this unit test to threshold cost test
<a href="https://github.com/sappenin"></img></a> **Issue by [sappenin](https://github.com/sappenin)**
_Saturday Oct 21, 2017 at 15:28 GMT_
_Originally opened as https://github.com/interledger/java-crypto-conditions/issues/78_

----

```
 For example, if a threshold crypto-condition contains 5 sub-
   conditions with costs of 64, 64, 82, 84 and 84 and has a threshold of
   3, the cost is equal to the sum of the largest three sub-condition
   costs (82 + 84 + 84 = 250) plus 1024 times the number of sub-
   conditions (1024 * 5 = 5120): 5370
```

pascalav:
* Add unit-tests covering the cost computation of a THRESHOLD-SHA-256 crypto-condition


crypto-conditions/src/test/java/org/interledger/cryptoconditions/ThresholdSha256ConditionTest.java
```diff
@@ -58,4 +58,49 @@ public void testConstructionUsingMultipleThreads() throws Exception {
     this.runConcurrent(1, runnableTest);
     this.runConcurrent(runnableTest);
   }
+
+  /**
+   * Tests the cost of a {@link ThresholdSha256Condition}. This test
+   * validates the enhancement for Github issue #10 where algorithm is
```
sappenin: This PR looks good, except the above comment doesn't make sense to me. I think we should strike the sentence, `This validates the enhancement for Github issue #10 where algorithm is reminded`, and instead add `<p>` tags so the overall Javadoc looks like this:

```
/**
 * <p>Tests the cost computation of a {@link ThresholdSha256Condition}.</p> 
 * 
 * <p>For example, if a threshold crypto-condition contains 5
 * sub-conditions with costs of 64, 84, 82, 64 and 84, and has a
 * threshold of 3, the cost is equal to the sum of the largest three
 * sub-condition costs (82 + 84 + 84 = 250) plus 1024 times the
 * number of sub-conditions (1024 * 5 = 5120): 5370.</p>
 *
 * @see "https://github.com/hyperledger/quilt/issues/10"
 * @see "https://github.com/interledger/java-crypto-conditions/issues/78"
 */
```

pascalav: Indeed. The PR brings this change in commit https://github.com/hyperledger/quilt/pull/70/commits/cf17438dc0dcaf8a0687bade0f450e9ec6985c5d.

adrianhopebailie:
@pascalav can you resolve the conflict here and we can merge?

pascalav:
@adrianhopebailie That should be fine now: I refreshed the PR branch from `hyperledger:master` and refactored it accordingly (crypto-conditions classes became immutables in the meantime).

I didn't get why `ThresholdSha256ConditionTest.java` conflicted, anyway.

So far, the commits tree is not clean for this PR. When it'll be approved, do you have the ability to squash it in 1 commit on `hyperledger:master`? (I hope so :p)

sappenin:
LGTM - this needs to be sync'd with master, but should be good to push.

PR #88, opened 2018-01-31
Add connector example folders

sappenin:
During the quilt call, we all agreed it would be ideal to locate various connector implementations _in_ the Quilt project. This commit provides some scaffolding to organize that effort.


adrianhopebailie:
Do we need a POM? Wouldn't each example connector have it's own?

sappenin:
I assume you're asking if [this pom](https://github.com/hyperledger/quilt/pull/88/files#diff-668257a561258a316a842876bf01c59e) is needed?

If yes, this pom just connects the main quilt-parent pom to the connector-example sub-modules. We could alternatively just link to each connector directly from the [main quilt project pom](https://github.com/hyperledger/quilt/pull/88/files#diff-600376dffeb79835ede4a0b285078036R92) if you want. If we do that, you'll see 1 module per connector example in the main quilt root pom. 

Otherwise, if we leave it as-is, you'll see [one pointer to the connector-examples parent pom](https://github.com/hyperledger/quilt/blob/daaa3bd3629f968011428bf02b815871c0338285/pom.xml#L95), and then in the connector-examples project, [you'll see a pom with all the connector sub-modules](https://github.com/hyperledger/quilt/pull/88/files#diff-668257a561258a316a842876bf01c59eR24) defined.

I'm open to changing, but I suggest we leave the pom's as-is. It will make the root pom.xml less cluttered.

earizon:
I'm thinking now that adding concrete implementation/examples/PoC to quilt will not be a good idea, because it will overflow the git quilt repository with spurious commits and also is to be expected that examples will not receive the same "care" that core quilt or get updated at the same pace. A developer can make "hundreds" of commits to a connector while deciding what's the best way to implement a feature. It will add lot of noise to keep them in quilt and I don't think it will also be friendly to the current CI. Maybe a simple table external_projects.md poiting to external git repos will be preferred (for example with columns for name, github url, description, last-update). For "important" external repos (reference implementations) git submodules can be used to easify developers work with them in a "relatively standard" way.

sappenin:
@earizon That makes a lot of sense to me, especially in the near-term when we're trying to explore best practices around creating a connector. I really like the idea of the `external_projects.md`.

@adrianhopebailie what do you think? If we all agree then I can cleanup what was merged by this PR.

adrianhopebailie:
I agree. Let's keep this code as un-opinionated as possible.

PR #89, opened 2018-01-31
Earizon pr85 review

earizon:
"Bugs" found during testing with java connector


ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -35,7 +35,7 @@
 public interface InterledgerAddress {
 
   String REGEX = "(?=^.{1,1023}$)"
-      + "^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$";
+      + "^(g|private|example|peer|self|test([1-3])?)[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$";
```
adrianhopebailie: 👍 - depends on https://github.com/interledger/rfcs/pull/376

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
```diff
@@ -24,7 +24,7 @@
    */
   static PreimageSha256Fulfillment from(final byte[] preimage) {
     Objects.requireNonNull(preimage);
-    final String encodedPreimage = Base64.getUrlEncoder().encodeToString(preimage);
+    final String encodedPreimage = Base64.getEncoder().encodeToString(preimage);
```
adrianhopebailie: We should decide if we are using Base64Url or Base64 and be consistent.

Personally I agree with @earizon that we should use Base64 as it is the most common.

ilp-core/src/main/java/org/interledger/core/InterledgerProtocolException.java
```diff
@@ -27,4 +27,10 @@ public InterledgerProtocolException(final InterledgerRejectPacket interledgerRej
   public InterledgerRejectPacket getInterledgerRejectPacket() {
     return interledgerRejectPacket;
   }
+
+  @Override
+  public String toString(){
+      return interledgerRejectPacket.toString();
+  }
```
adrianhopebailie: Should also print the result of `getMessage()`?

ilp-core/src/main/java/org/interledger/core/InterledgerRejectPacket.java
```diff
@@ -4,6 +4,10 @@
 
 import java.util.Arrays;
 
+/**
+ * <p>REF: https://github.com/interledger/rfcs/blob/master/asn1/InterledgerProtocol.asn</p>
+ */
```
adrianhopebailie: Rather use: https://interledger.org/rfcs/asn1/InterledgerProtocol.asn.html

adrianhopebailie:
@earizon I have addressed some comments from @sappenin 
Can you rebase these and I'll merge them in before merging everything into master

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
sappenin: This change isn't enough to make this work -- we would also need to change the `PREFIX_REGEX`. I made this change (and added unit tests) in a new PR #90. 

I suggest we remove this change from this PR, and use #90 instead.

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
sappenin: @earizon did something happen in your test code to prompt this change? I guess I'm not intrinsically opposed to the change (using Base64), although in my mind it feels like defaulting to Base64Url is better for anyone who might want to put a String into a URL -- so I find myself always defaulting to that instead.

That said, I'm not aware of any use-cases where somebody would want to put the encoded preimage from a Fulfillment into a URL, so in this case, it's probably a non-issue, and moving to Base64 isn't actually a problem. I suppose I'm more interested in what prompted the change.

In any event, this change isn't sufficient because we would need to change the encoding in other places in the project (e.g., see [CryptoConditionWriter#writeTypedFulfillment](https://github.com/hyperledger/quilt/blob/master/crypto-conditions/src/main/java/org/interledger/cryptoconditions/CryptoConditionWriter.java#L159)).

I suggest we pull this change out of this PR, and address it in a different PR because it seems like a bigger change that we would want to vet on its own.

earizon: @adrianhopebailie , @sappenin Actually it was a request from @michielbdejong to make it compatible with testnet-of-testnets and ilp-over-http.

ilp-core/src/main/java/org/interledger/core/InterledgerProtocolException.java
earizon: @adrianhopebailie This comes for free in InterledgerRejectPacket::toString defined as:

          return "InterledgerRejectPacket{"
          + "  code=" + getCode()
          + ",  triggeredBy=" + getTriggeredBy()
          + ",  message=" + getMessage()
          + ",  data=" + Arrays.toString(getData())
          + '}';

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
earizon: @sappenin  I agree. #90 fixes it properly. 

I don't think I can remove from the PR (or can I?) but it will be immediate to fix during merge-conflict-resolution.

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
sappenin: Ah, cool. @michielbdejong Can you comment here? 

The property in question is just the encoded-preimage of a `PreimageSha256Fulfillment` --
 it's not the OER bytes of the PreimageSha256Fulfillment. Instead, it's literally just a helper-method to access the bytes of the preimage (not the bytes of the Fulfillment itself).

So, I'm surprised that anybody is doing anything with this particular field (enough to care about this encoding). Any thoughts?

sappenin: Also, I've created https://github.com/hyperledger/quilt/issues/92 to correct the Javadoc of this method.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
sappenin: Either way is fine - I think you can just remove this change in your branch, and then push another commit (which will just restore this class to its previous state). 

We probably want to do this for the crypto-condition change above as well since these two things have quite a bit of discussion, and are artificially blocking your other changes.

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
earizon: @sappenin What's the difference between the bytes of the preimage and the bytes of the Fulfillment?

I'm using it to return the [ilp-over-http](https://github.com/interledger/rfcs/blob/ilp-over-http/0000-ilp-over-http.md) `ILP-Fulfillment` header.

adrianhopebailie: We should separate the needs of a specific application or protocol (like `ilp-over-http`) from the design of this library which could be a dependency in a variety of projects.

The method name `getEncodedPreimage` doesn't provide any hints as to whether it's Base64 or Base64Url but the Javadoc does say Base64Url so perhaps we should leave it as is.

Remember that the motivation for this was to return a String and not a `byte[]` to avoid mutability issues. What we are really returning is 32-bytes of data but in a form that's immutable so the encoding is not important as long as it's consistent and not surprising to a user.

ilp-core/src/main/java/org/interledger/core/InterledgerProtocolException.java
adrianhopebailie: I mean `InterledgerProtocolException::getMessage()` not `InterledgerProtocolException::getInterledgerRejectPacket()::getMessage()`

adrianhopebailie:
@earizon I am going to merge  `refactor/ilpv4` into master and then you can change this PR to merge into `master` when it's ready

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
sappenin: @earizon One set of array of bytes is the bytes of the preimage, and the other array of bytes is the bytes of the whole Fulfillment. For example: 

* **Bytes of the preimage**: "secret preimage".getBytes()
* **Bytes of the Fulfillment**: CryptoConditionWriter.write(fulfillment);

The first one is the literal bytes of whatever is the preimage. The second one is the OER-encoded bytes of the entire fulfillment, encoded as whatever you wish (probably in the test-net, this would be: a bunch of DER bytes, encoded in Base64).

I think that's where we're missing each other. Checkout the actual bytes that get written when you write an entire fulfillment in [CryptoConditionWriter.java](https://github.com/hyperledger/quilt/blob/master/crypto-conditions/src/main/java/org/interledger/cryptoconditions/CryptoConditionWriter.java#L85).

In the [ilp-over-http](https://github.com/interledger/rfcs/blob/ilp-over-http/0000-ilp-over-http.md) I think you want to return the bytes of the entire fulfillment, like this:

```
byte[] fulfillmentBytes = CryptoConditionWriter.write(fulfillment);
return Base64.getEncoder().getEncode(fulfillmentBytes);
```

sappenin: > The method name getEncodedPreimage doesn't provide any hints as to whether it's Base64 or Base64Url but the Javadoc does say Base64Url so perhaps we should leave it as is.

See #95 for tracking this one - I'm open to method-name-change suggestions. At the least, we need to cleanup the Javadoc for that method, and probably mention that it's purpose is strictly for immutability, not encoding the entire fulfillment.

adrianhopebailie:
Creating new PR to merge into `master`

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
sappenin: @earizon Oops, my comment above about the difference in encoding (and the encoding you should be using) is _not_ correct for your use-case. 

While my comment is correct from the perspective of the crypto-conditions RFC, [ilp-over-http](https://github.com/interledger/rfcs/blob/ilp-over-http/0000-ilp-over-http.md#request) encodes conditions/fulfillments using a different encoding, which I forgot about.

(My apologies for that oversight).

So, all that to say...you can ignore my comment above where I outline the difference between preimage and fulfillment bytes....in your case, I think you'd want to do one of the following:

```java
PreimageSha256Fulfillment fulfillment = ...;
byte[] preimageBytes = Base64.getUrlDecoder().decode(fulfillment.getEncodedPreimage());
String base64Preimage = Base64.getEncoder().encodeToString(preimageBytes);
```

It's ugly, I know - but let's continue discussing a change to the `crypto-conditions` project in #100.

PR #83, opened 2018-01-25
fix: allow Interledger testnet addresses

michielbdejong:
Interledger testnet addresses all start with `test.`.

PR #90, opened 2018-02-02
Fixes #83

sappenin:
https://github.com/interledger/rfcs/pull/376 allows for a slightly different prefix of just “test.” in an ILP address. This PR adjusts the allowed prefixes to conform to this change.

Also adds unit test coverage.


adrianhopebailie:
👍 - can you merge into #85 too?

ilp-core/pom.xml
```diff
@@ -10,7 +10,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.interledger</groupId>
     <artifactId>ilp-core</artifactId>
-    <version>0.9.2-SNAPSHOT</version>
```
earizon: Constantly changing SNAPSHOT version can just cause trouble.

Some code can inadvertently work for long time with a old 0.x.x-SNAPSHOT just because we forgot to update the dependency and the outdate artifact was still available in a remote repository. Then suddenly everything fails when changing the dependency version or the old artifact dissapears from the repo :(

I think it's just simpler (and less buggy) to just default to `1.0.0-SNAPSHOT`. -SNAPSHOT suffix is designed to reflect the intention to release a future (1.0.0) version. This way all software depending on the snapshot will stop compiling when a new 1.0.0-SNAPSHOT is published (and that's what, most probably, everybody wants). Changing the minor version just indicates the intention to release a future stable 0.9... version, then a future stable 0.10.... version, and so on. But I do not think this is the intention right now. Since those version will be automatically deprecated by any new SNAPSHOT release.

The common "acepted-by-most-developpers" way of working is similar to:
  - During development of a new version, feature enhacement or bug fix point to  `MajorVersion.MinorVersion.IncrementalVersion-SNAPSHOT` future realease.
  - At release time remove -SNAPSHOT. (future to present)
  - For bug-fix after release (non SNAPHOT that can be used by production enviroments)  point to `MajorVersion.MinorVersion.++IncrementalVersion-SNAPSHOT`
  - During feature enhancement point to release version `version.++MinorVersion.0 (-SNAPSHOT)`
  - For breaking API point to `++MajorVersion.0.0-SNAPSHOT`

[This blog](https://www.dontpanicblog.co.uk/2011/12/13/maven-release-plugin/)  contains some more info about how maven-release-plugin expect versions to be handled.

sappenin: @earizon the process you propose is how I see things working once we have a `1.0.x` release. 

For now, we should either have _real_ dot-releases, or else keep the current system of just incrementing the SNAPSHOT builds. If we have consensus around creating dot-releases (non-SNAPSHOTs) then I don't mind doing that.

But currently, we're not releasing pre-1.0 _release_ artifacts (only SNAPSHOTS), so the reason we _MUST_ increment the SNAPSHOT versions is because we expressly _do not_ want to force any developer using a SNAPSHOT to update code. We want to give developers the choice -- either keep using the outdated SNAPSHOT, or update to a newer SNAPSHOT.

To your point about missing SNAPSHOTs, have you encountered this issue with Quilt artifacts?
 I'm only publishing these once to Sonatype, so there should never be a time (using the current system) that code stops working due to a misssing or changed SNAPSHOT. If that's not the case, we should figure out why.

I guess we need to take a quick vote before we push or adjust this PR:

- [ ] Keep the current system (update SNAPSHOT every time we make a change).
- [ ] Create an actual pre-1.0 release everytime we change something (e.g., `0.9.2`, `0.10.0`, etc).
- [ ] Just use `1.0.0-SNAPSHOT` and break anyone using older SNAPSHOT code.

I vote to keep the current system. @adrianhopebailie @earizon or anyone else?

earizon: @sappenin "We want to give developers the choice -- either keep using the outdated SNAPSHOT, or update to a newer SNAPSHOT"
Why giving the choice to do the non-standard thing? Git tags are much better suited for that. If a developer just wants to use an outdated image (in my experience that's not a good idea most of the times) all he has to do is to "git checkout" to the given tag and install  locally (dev-ops/util/quick_local_install_mvn_package.sh allows to compile in a couple of seconds bypassing tests, format,...).

"have you encountered this issue with Quilt artifacts?"
At this moment any project dependent of Quilt  will need to constantly check what's in the quilt pom and then update its pom dependencies just to be able to use the latest published artifact. Using a constant 1.0.0-SNAPSHOT fix de problem. An -alpha- | -beta- | -pre- is commonly used to warn that maybe 1.0.0-alpha-SNAPSHOT is far away from stable.

adrianhopebailie:
Moved release process and versioning discussion to #95

adrianhopebailie:
@sappenin this needs to be rebased after merging in https://github.com/hyperledger/quilt/pull/85

sappenin:
@earizon I'm blocked from merging until you approve. Does #95 address your concern? If so, can you remove your block or approve?

sappenin:
> @sappenin this needs to be rebased after merging in #85

This is all set and ready to merge...

earizon:
@sappenin  Upps, sorry, my e-mail is overloaded with millions of github mails. I just noticed it 5 days later.

sappenin:
Adds extra unit tests for #83. Also, some small cleanup.

PR #99, opened 2018-02-06
Earizon pr85 review

adrianhopebailie:
"Bugs" found during testing with java connector

Reopened from #89 to target`master`

### DIFF
--- a/crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
+++ b/crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
@@ -24,7 +24,7 @@ public interface PreimageSha256Fulfillment extends Fulfillment<PreimageSha256Con
    */
   static PreimageSha256Fulfillment from(final byte[] preimage) {
     Objects.requireNonNull(preimage);
-    final String encodedPreimage = Base64.getUrlEncoder().encodeToString(preimage);
+    final String encodedPreimage = Base64.getEncoder().encodeToString(preimage);
 
     final long cost = AbstractPreimageSha256Fulfillment.calculateCost(preimage);
     final byte[] fingerprint = HashUtils.hashFingerprintContents(
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 public interface InterledgerAddress {
 
   String REGEX = "(?=^.{1,1023}$)"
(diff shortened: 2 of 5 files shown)


crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
```diff
@@ -24,7 +24,7 @@
    */
   static PreimageSha256Fulfillment from(final byte[] preimage) {
     Objects.requireNonNull(preimage);
-    final String encodedPreimage = Base64.getUrlEncoder().encodeToString(preimage);
+    final String encodedPreimage = Base64.getEncoder().encodeToString(preimage);
```
adrianhopebailie: As discussed in #89 I think we should revert this.

If an external application needs the Base64 (not Bse64Url) encoded form it should convert between them

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -35,7 +35,7 @@
 public interface InterledgerAddress {
 
   String REGEX = "(?=^.{1,1023}$)"
-      + "^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$";
+      + "^(g|private|example|peer|self|test([1-3])?)[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$";
```
adrianhopebailie: We can revert this too as it's addressed in #90

ilp-core/src/main/java/org/interledger/core/InterledgerProtocolException.java
```diff
@@ -27,4 +27,10 @@ public InterledgerProtocolException(final InterledgerRejectPacket interledgerRej
   public InterledgerRejectPacket getInterledgerRejectPacket() {
     return interledgerRejectPacket;
   }
+
+  @Override
+  public String toString(){
+      return interledgerRejectPacket.toString();
+  }
+
 }
```
adrianhopebailie: As discussed in #89 I think this should include the value of `this.getMessage()`

crypto-conditions/src/main/java/org/interledger/cryptoconditions/PreimageSha256Fulfillment.java
sappenin: Agree. Let's revert for now and discuss in #100.

sappenin:
Closing this for now - as discussed on the last call, there's only one change in here that we actually want to keep, but it probably makes sense to open a new PR for that. I've opened #134 to track that and assigned it to @earizon.

PR #102, opened 2018-02-07
Fixes #57

### LINKED ISSUE #57: [ilp-core] Enforce Regexes from RFC-15
From [RFC-0015](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#address-requirements)

>  5.) Destination addresses [...] MUST contain at least two segments after the allocation scheme prefix.

Currently, `InterledgerAddress.java` doesn't enforce that a destination address have at least 2 segements (i.e., `g.foo` is allowed in the `InterledgerAddress` implementation, but should not be).

We should update `#isValidInterledgerAddress` to utlilize a different regex depending on whether or not the String ends with a period [.]. Additionally, we might consider improving the error output instead of just saying, "reference ILP-RFC-15."

pascalav:
* Enforce the RFC-15 regular expressions for address prefixes and destination addresses respectively


pascalav:
Hi, here are some additional comments about the PR:
* Mainly, the 2 REGEX for address prefix and destination address got enforced in `InterledgerAddress#isValid(java.lang.String)`, and unit tests got refactored accordingly (i.e. tests that were using invalid 1-segment destination addresses - such as `g.foo` - are now using valid >=2-segments destination addresses - such as `g.foo.bar` or `g.foo.bar.baz` -)
* A new unit test was added to ensure a reject of 1-segment destination addresses 
* Critical modification: `InterledgerAddress#getParentPrefix()` algorithm was changed, in order to make it possible to call `InterledgerAddress.of("g.foo.").getParentPrefix().get().getValue() // expected: "g."`. This because a destination address smallest shape is `<scheme>.<segment_1>.<segment_2>`, and an address prefix smallest shape is `<scheme>.`. But in the current code base, `g.foo. --> g.` is walked-up through the `g.foo. --> g.foo --> g.` path, which breaks on the (in)validated-and-immutable `g.foo` destination address that is in the middle

About issuing a more precise error message when the address is invalid, I think it would better suit into another evolution ticket, since this refactoring already has a large foot print in the code base. If you're okay with that, I'll open a new ticket, holding a reference on https://github.com/hyperledger/quilt/issues/57 .

adrianhopebailie:
> I think it would better suit into another evolution ticket

👍

sappenin:
Nice work @pascalav!

pascalav:
You're both welcome :+1: 

@sappenin It's ok, I'll write explicit constant names in future contributions.

About more explicit error messages upon an `InterledgerAddress#check()` reject, I opened https://github.com/hyperledger/quilt/issues/104. Let me know if it's not so good for visitors to open issues: I believe one can't add labels to them when having that role (i.e. tagging as `evolution`, `winter games`, `foo`).

PR #105, opened 2018-02-08
Node Framework

adrianhopebailie:
A framework for new ILP nodes.
Lots of features and tests to still add.

TestNode provides a clue as to how to use it.
Next step is to see if it is framework-agnostic enough to be used within Spring and VertX.


ilp-node/src/main/java/org/interledger/node/channels/Channel.java
```diff
@@ -0,0 +1,182 @@
...
+   * Sends an ILP request packet to the peer and returns the response packet.
+   *
+   * @param request The request to send to the peer.
+   *
+   * @throws ChannelNotOpenException if the channels is not connected.
+   * @throws InterledgerProtocolException if the request is rejected by the peer.
+   *
+   * @return A Future that resolves to the ILP response from the peer.
+   */
+  Future<InterledgerFulfillPacket> sendRequest(InterledgerPreparePacket request);
```
sappenin: I think I prefer `sendPacket` instead -- it's surprising to see a method called "sendRequest" and then see that I don't actually do anything with an object that refers to a "request" anywhere.

ilp-node/src/main/java/org/interledger/node/channels/Channel.java
```diff
@@ -0,0 +1,182 @@
...
+   * <p>Example: If this channels is used to make transfers over a payment channel then this should
+   * result in a new claim being sent over the channel.
+   *
+   * @param amount The amount to credit the account in favour of the sender.
+   *
+   * @throws ChannelNotOpenException if the channel is not open.
+   *
+   * @return a Future that resolves when the transfer has been sent.
+   */
+  Future<Void> sendTransfer(long amount);
```
sappenin: The naming of this and `sendRequest` are a little confusing/surprising to me, especially after having looked at the [LPI2](https://interledger.org/rfcs/0024-ledger-plugin-interface-2/) for so long.

This operation doesn't _quite_ feel like a transfer (I could be wrong there) -- though I don't really like `sendMoney` either. 

I don't have a good replacement name here, but am wondering what this method is actually doing? I mean, conceptually, "sending money" or even "transferring money from one account to another" makes sense to me (though I would like to be less currency-focused in ILP, so don't like the term `money`). But is this more like clearing or settlement? Or something else? I don't think we should use those terms either (clearing,settle, etc), but am instead trying to figure out -- in layman's terms -- what this operation _does_.

Is it just transferring value from one account to another, across a channel?

adrianhopebailie: The node should call this when it has made an adjustment to the balances it is keeping for the account. So imagine the following flow:

  1. Node sends multiple ILP Prepare packets down the Channel to a total value of X.
  2. Balance on the Account for that Channel is now -X.
  3. Node wants to bring the balance back to 0 so it calls `sendTransfer(X)` and adjusts the balance to 0.

As discussed on the call I think we can model this better by separating these functions. I'm going to have an attempt at that today.

ilp-node/src/main/java/org/interledger/node/channels/Channel.java
adrianhopebailie: This is interesting because when I discussed this with @emschwartz we concluded that an ILP Prepare is a request and an ILP Fulfill is a response.

The LPIv2 model is to treat this as raw data and the plugins assume it can be serialized/deserialized into an ILP packet.

I have a thought to wrapping this in a Request/Response which for now might only have the ILP packet but in future might contain the raw request bytes to allow the nodes to have very fast in-place replacement of data rather than a deserialization and then serialization step.

sappenin: Makes sense - this is fine for now - maybe in future have two methods - one that takes the packet (like you have now) and one takes a request with bytes?

adrianhopebailie:
I am going to close this PR but leave the branch open for now as a reference.

As discussed on the call on 1 March we'll change this to have reference implementations of some services but less opinion on how to wire it all together.

PR #123, opened 2018-02-25
Fixes #104

### LINKED ISSUE #104: Improve the error output in `InterledgerAddress#check()`
As suggested by @sappenin , regarding enforcement of RFC-15 in `InterledgerAddress#check()`: 

> we might consider improving the error output instead of just saying, "reference ILP-RFC-15."

Reference: https://github.com/hyperledger/quilt/issues/57#issue-277115537

pascalav:
* Enhances the ILP address error check into a parser class

### DIFF
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -305,7 +305,11 @@ public interface InterledgerAddress {
      */
     @Value.Check
     void check() {
-      InterledgerAddress.isValid(getValue());
+      try {
+        InterledgerAddress.isValid(getValue());
+      } catch (final IllegalArgumentException e) {
+        throw new IllegalArgumentException("Address is invalid", e);
+      }
     }
   }
 
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -6,7 +6,6 @@ import org.immutables.value.Value;
 
 import java.util.Objects;
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 /**
  * Interledger Protocol (ILP) Addresses identify Ledger accounts (or groups of Ledger accounts) in
@@ -32,16 +31,7 @@ import java.util.regex.Pattern;
  */
 public interface InterledgerAddress {
 
-  String ADDRESS_PREFIX_REGEX = "(?=^.{1,1023}$)"
-      + "^(g|private|example|peer|self|test[1-3]?)[.]([a-zA-Z0-9_~-]+[.])*$";
-  Pattern ADDRESS_PREFIX_PATTERN = Pattern.compile(ADDRESS_PREFIX_REGEX);
-
-  String DESTINATION_ADDRESS_REGEX = "(?=^.{1,1023}$)"
-      + "^(g|private|example|peer|self|test[1-3]?)[.]([a-zA-Z0-9_~-]+[.])+[a-zA-Z0-9_~-]+$";
-  Pattern DESTINATION_ADDRESS_PATTERN = Pattern.compile(DESTINATION_ADDRESS_REGEX);
-
-  String ROOT_REGEX = "^(g|private|example|peer|self|test[1-3]?)[.]$";
-  Pattern ROOT_PATTERN = Pattern.compile(ROOT_REGEX);
+  InterledgerAddressParser ADDRESS_PARSER = new InterledgerAddressParser();
 
   /**
    * Constructor to allow quick construction from a String representation of an ILP address.
@@ -67,9 +57,8 @@ public interface InterledgerAddress {
    */
   static boolean isValid(final String value) {
     Objects.requireNonNull(value);
-    return (value.endsWith(".") ? ADDRESS_PREFIX_PATTERN : DESTINATION_ADDRESS_PATTERN)
-        .matcher(value)
-        .matches();
+    ADDRESS_PARSER.validate(value);
+    return true;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


ilp-core/src/test/java/org/interledger/core/InterledgerAddressBuilderTest.java
```diff
@@ -72,46 +69,22 @@ public void testConstruction_LedgerPrefix() {
...
+    InterledgerAddress.builder().value("").build();
   }
 
   @Test(expected = IllegalArgumentException.class)
   public void test_blank_address() {
-    final String value = "  ";
-    try {
-      InterledgerAddress.builder().value(value).build();
-    } catch (IllegalArgumentException e) {
-      assertThat(e.getMessage(), is(String.format(EXPECTED_ERROR_MESSAGE, value)));
```
sappenin: I think we should keep the error message assertions. Primarily, it validates that the `IllegalArgumentException` that was thrown was actually what we expected, as opposed to being some _other_ IllegalArgumentException. 

Thoughts?

sappenin:
Overall, these changes LGTM -- one blocking comment, but once that's resolved this seems mergeable.

ilp-core/src/test/java/org/interledger/core/InterledgerAddressBuilderTest.java
pascalav: I don't think so. Here is my point of view, through another domain use-case:

`doorKeeper#canStartBlackFriday` // possible catches: `goodsUnavailableException(pendingDeliveryException)`, `goodsUnavailableException(supplierOutOfStockException)`, `tooEarlyException("Disallowed before 9AM")`
__-> `warehouseManager#areGoodsAvailable` // possible catches: `pendingDeliveryException, supplierOutOfStockException`
____-> `transportPartner#isOrderDelivered`
____-> `supplier#isSerialNumberInStock`
__-> `timeService#isGTE9AM`

By separating concerns, the `doorKeeper` may:
* Either, delay the doors opening until all lights are green
* Either, contact the `storeOwner` and forward him the underlying `goodsUnavailableException` or `tooEarlyException` for decision
* Either, delay the doors opening in case of a `goodsUnavailableException`
* Either, bypass restriction with a tolerance threshold in case of a `tooEarlyException`

That is, he has a shallow view of the situation, and should probably not be given a deeper visibility on nested events (i.e. knowing that goods are not available due to a supplier shortage or to a transport issue is too much knowledge for his business imo).

Put back in our situation, PR is currently:

`test-caller` // catches: `illegalArgumentException("Address has no scheme prefix")`
__->addressBuilder#build // passes through the unaltered: `illegalArgumentException("Address has no scheme prefix")`
____-> `addressParser#validate`

Just now, I would change this into:

`test-caller` // catches: `illegalArgumentException("Address is invalid", illegalArgumentException)` (and we may imagine other causes such as `"Address has no route to destination"` even if it seems unsuitable to be managed so early at runtime)
__-> `addressBuilder#build` // passes through: `illegalArgumentException("Address has no scheme prefix")`
____-> \_ `addressParser#validate`

And the `addressBuilder` test-class would check for the `Address is invalid` error message, rather than digging into the parser internals (which are already checked in the parser test-class, therefore would be redundant in here).

In the long run, rather than relying on error messages only, I would subclass `InterledgerRuntimeException` into `InvalidAddressException`, `AnotherCauseException` and so on.
`test-caller` would then shallow-check the exception type only.

ilp-core/src/main/java/org/interledger/core/InterledgerAddressParser.java
```diff
@@ -0,0 +1,133 @@
...
+  private static final Pattern SCHEME_PREFIX_ONLY_PATTERN = Pattern.compile(
+      SCHEME_PREFIX_ONLY_REGEX
+  );
+
+  /**
+   * Validates an ILP address.
+   * 
+   * @param addressString The ILP address to validate
+   */
+  void validate(final String addressString) {
```
pascalav: Do you think that the method signature - as well as its Javadoc - should state `throws IllegalArgumentException`?

I think so, even if mentionning unchecked exceptions in method signatures is bad practice (because most of time, they are not checked/catched by caller). This way, a dev doesn't have to dig into the method's logic to finally find that validation gets rejected through a `RuntimeException`.

ilp-core/src/test/java/org/interledger/core/InterledgerAddressBuilderTest.java
pascalav: Anyway, I can do as you suggested (which also ends up in redundancy of the address-invalidity cause check among test-classes), just confirm and I'll do so.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -67,9 +57,8 @@ static InterledgerAddress of(final String value) {
    */
   static boolean isValid(final String value) {
     Objects.requireNonNull(value);
-    return (value.endsWith(".") ? ADDRESS_PREFIX_PATTERN : DESTINATION_ADDRESS_PATTERN)
-        .matcher(value)
-        .matches();
+    ADDRESS_PARSER.validate(value);
```
pascalav: To illustrate the discussion about the "cause of `IllegalArgumentException`" check in the `InterledgerAddressBuilderTest` class, I would change this line into:

> try {
>   ADDRESS_PARSER.validate(value);
> } catch (final IllegalArgumentException e) {
>   throw new IllegalArgumentException("Address is invalid", e)
> }

Then `InterledgerAddressBuilderTest` would only rely on that later message.

ilp-core/src/main/java/org/interledger/core/InterledgerAddressParser.java
sappenin: Sounds fine to me.

pascalav: This point is now adressed.

ilp-core/src/test/java/org/interledger/core/InterledgerAddressBuilderTest.java
pascalav: This point is now adressed: I restored the `IllegalArgumentException` checks in test-classes that use the `InterledgerAddressBuilder`.

PR #127, opened 2018-03-12
Feature/#56 allow overridden error message in interledger address checks

pascalav:


ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -81,15 +81,46 @@ static boolean isValid(final String value) {
...
+   *
+   * @param address A {@link InterledgerAddress} to check.
+   * @param errorMessage An error message to output upon invalidation.
+   *
+   * @return {@code ledgerPrefix} if its value ends with a dot (.).
+   *
+   * @throws IllegalArgumentException if the supplied Interledger address is not a
+   *     ledger-prefix. Embeds the {@code errorMessage}.
+   */
+  static InterledgerAddress requireAddressPrefix(final InterledgerAddress address,
```
sappenin: One minor nit with this implementation -- it allows a caller to make the following call without an error:

```
InterledgerAddress.requireAddressPrefix(address, null);
```

This should throw an error, because we don't want our public API contracts to be operating on [null values](https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained).

This can be rectified by inverting the calls between the two methods. So, if `requireAddressPrefix(address)` is called, it should internally call `requireAddressPrefix(address, "addressPrefix must not be null!")`. Then we can add an `Objects.requireNotNull` check inside of `requireAddressPrefix(address, message)` on line 115 for the message.

pascalav: I'm gonna refactor accordingly.

However, there are 2 default error messages in `#requireAddressPrefix(InterledgerAddress)` currently: `"InterledgerAddress must not be null!"` and `String.format("InterledgerAddress '%s' must be an Address Prefix ending with a dot (.)",address.getValue())`.

As a consequence, a `private` `#checkAddressPrefix(InterledgerAddress, String, String)` will probably be needed in order for `#requireAddressPrefix(InterledgerAddress, String)` to comply with the error messages multiplicity.

pascalav: Finally I did not find a simple refactor in order to put the overriden message in place of defaults. 

Made this through the `InterledgerAddressParser`, since the `InterledgerAddress` interface cannot have both `private` and `static` methods. I didn't want to put additional weigh on that facade class also.

sappenin: Ah, yes, seeing that now. OK, LGTM!

PR #131, opened 2018-03-18
Fixed pom versioning

dora-gt:
- mvn build fails because of pom setting, version mismatch.
  - `dev-ops` is `0.1.1-SNAPSHOT` but specified `0.1.0-SNAPSHOT` in `pom.xml`
  - `ilp-annotations ` is `0.1.1-SNAPSHOT` but specified `0.1.0-SNAPSHOT` in `pom.xml` and `ilp-annotations/pom.xml`

### DIFF
--- a/ilp-annotations/pom.xml
+++ b/ilp-annotations/pom.xml
@@ -11,7 +11,7 @@
   <name>Interledger Annotations</name>
   <groupId>org.interledger</groupId>
   <artifactId>ilp-annotations</artifactId>
-  <version>0.1.0-SNAPSHOT</version>
+  <version>0.1.1-SNAPSHOT</version>
   <modelVersion>4.0.0</modelVersion>
 
   <packaging>jar</packaging>
--- a/ilp-core-codecs/README.md
+++ b/ilp-core-codecs/README.md
@@ -24,7 +24,6 @@ final InterledgerPreparePacket packet
   .executionCondition(
       PreimageSha256Condition.fromCostAndFingerprint(32, preimage))
   .expiresAt(Instant.now().plus().plusSeconds(30))
-  .data(new byte[] {})
   .build();
 
 context.write(packet, outputStream);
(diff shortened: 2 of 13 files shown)


sappenin:
Let's align the versioning of all the modules so we don't encounter this issue in the future. To do that, we'll need to do the following:

1. Update the parent pom.xml to have version `0.11.0-SNAPSHOT`.
1. In each of `ilp-core`, `ilp-annotations`, `dev-ops`, and `codec-framework`, remove the `<group-id>` and the `<version>` tags so that each module simply inherits these from the parent pom.

For example, `ilp-core` would look like this:

```
  <parent>
    <artifactId>quilt-parent</artifactId>
    <version>0.1.0-SNAPSHOT</version>
  </parent>
  <modelVersion>4.0.0</modelVersion>
  
  <artifactId>ilp-core</artifactId>
  
  <name>Interledger Core</name>
``` 

Finally, in the `<dependencyManagement>` section in the parent pom, we will need to update the versions to be as follows:

```
 <dependency>
        <groupId>${project.groupId}</groupId>
        <artifactId>ilp-annotations</artifactId>
        <version>${project.version}</version>
      </dependency>

      <dependency>
        <groupId>${project.groupId}</groupId>
        <artifactId>codec-framework</artifactId>
        <version>${project.version}</version>
      </dependency>

      <dependency>
        <groupId>${project.groupId}</groupId>
        <artifactId>ilp-core</artifactId>
        <version>${project.version}</version>
      </dependency>

      <dependency>
        <groupId>${project.groupId}</groupId>
        <artifactId>ilp-dev-ops</artifactId>
        <version>${project.version}</version>
      </dependency>
```

**_Note two things_**:

1. We'll need to update the `dev-ops` project to inherit from `quilt-parent`.
1. Let's not touch anything related to the `crypto-conditions` modules -- I'm working on a PR to extract these from the project into [https://github.com/hyperledger/quilt-crypto-conditions](https://github.com/hyperledger/quilt-crypto-conditions).

adrianhopebailie:
Should we move crypto-conditions into it's own project first?

sappenin:
> Should we move crypto-conditions into it's own project first?

CryptoConditions is pretty independent as far as versioning is concerned - we could leave that along, and just update everything else.

adrianhopebailie:
I've fixed this in #138 
All submodules inherit from the parent per Maven convention
Also added the `versions` plugin to help with managing versions: http://www.mojohaus.org/versions-maven-plugin/

adrianhopebailie:
Thanks @dora-gt for pushing this along

PR #146, opened 2018-05-18
Fixes #135 : Ready for review

### LINKED ISSUE #135: Resolve licensing issues discovered by Fossology scan
## Context
Hyperledger Quilt source files (from April 3, 2018) were scanned by The Linux Foundation utilizing Fossology to determine license for files contained in the following repositories:
  * https://github.com/hyperledger/quilt
  * https://github.com/hyperledger/quilt-crypto-conditions

From the Hyperledger Charter (https://www.hyperledger.org/about/charter) Section 13, all files that are included in Hyperledger projects should either be licensed with the Apache License, Version 2.0 for source code or Creative Commons Attribution 4.0 International License for documentation.

## Expected Behavior
All files should have license text included either directly or via a LICENSE file in the directory for files that cannot be modified to include a comment with the license text (think binary files).

## Actual Behavior
See results of scan: [quilt-2018-04.xlsx](https://github.com/hyperledger/quilt/files/1881025/quilt-2018-04.xlsx)

## Possible Fix
_ALL_ files that do not have a license included, the first step is to check with the contributor to ensure that the file can be released as Apache 2.0 or CC-BY-4.0.  If the contributor agrees, the file should be modified to include the license. You can utilize [SPDX Identifiers](https://spdx.org/) to represent the correct license.

For those non-modifiable files, you will need to include a LICENSE file in the directory containing the license that the files are covered by.  Here are a few examples of how this might be done:
* [OPNFV Apex](https://github.com/opnfv/apex/blob/master/LICENSE.rst) -- this is an example of explicitly calling out Apache 2.0 for source code and CC-BY-4.0 for docs.  
* [containerd](https://github.com/containerd/containerd/blob/master/README.md) -- see notice at bottom.  This also calls out the separate Apache / CC-BY-4.0 docs, and explicitly references the "docs" directory.

## Steps to Reproduce
Run code through Fossology (contact tkuhrt at linuxfoundation dot org) to have the source re-run after the modifications.

Gregvh:
The command mentioned in issue #135 does not work unless I add an organization name to the pom. This name will appear in the license headers. What name should I use ?

Also the dev-ops module has no reference to the parent pom, is it OK if I add it so that it inherits the organization name and inceptionYear (both needed by the plugin) ?

Thanks in advance for the feedback !


Gregvh:
@adrianhopebailie
Could you help me answer the questions in my previous comment ?
Thanks in advance !

sappenin:
@Gregvh for the organization name, let's just use `Interledger`

adrianhopebailie:
@Gregvh thanks for this, apologies for the delay in responding.

@tkuhrt does the copyright need to be Hyperledger, Linux Foundation or is Interledger okay?

tkuhrt:
See https://wiki.hyperledger.org/community/copyright-and-license-policy. It says, 

> for new files added to Hyperledger repositories, the following copyright text should be included:
> 
> `Copyright Hyperledger and its contributors.`

dev-ops/pom.xml
```diff
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <parent>
```
adrianhopebailie: I believe this is what is causing the build to fail on CI.

sappenin:
I don't think we can make the `dev-ops` module extend from quilt-parent because it's a dependency of `quilt-parent`.

This also appears to be why the build is breaking. I would say if we want to try to adjust the pom structure we should do that as part of a separate unit of work, independent of the licensing changes.

PR #159, opened 2018-07-13
Fix typos

sublimator:


sappenin:
LGTM, but when you commit, you need to add a DCO sign-off. It's pretty annoying because if you get even the slightest thing wrong, the system will reject it. See [DCO](https://github.com/probot/dco#how-it-works) for more details.

sublimator:
Hey, yeah, will fix

Meanwhile tron has 6000 commits: https://github.com/tronprotocol/java-tron

Started with network requests inside static init {} blocks

I often wonder if all this ceremony just gets in the way :/

sublimator:
@sappenin 

I force pushed with an amended -s commit

sappenin:
> I force pushed with an amended -s commit
Awesome! Last time I tried that the DCO bot wouldn't turn green, so glad to hear that's working now.

PR #163, opened 2018-07-22
Fixes #162

### LINKED ISSUE #162: VarUInt with value zero is encoded incorrectly
Bug is here:
https://github.com/hyperledger/quilt/blob/master/codec-framework/src/main/java/org/interledger/encoding/asn/codecs/AsnUintCodec.java#L52

This needs to exclude the case where the value == 0

```java
if (bytes[0] == 0x00 && bytes.length > 1) {
```

sappenin:
* Only strip leading bytes if the byte-length is < 1
* Add unit test for AsnUintCodec


codec-framework/src/main/java/org/interledger/encoding/asn/codecs/AsnUintCodec.java
```diff
@@ -38,18 +38,19 @@ public BigInteger decode() {
...
     byte[] bytes = value.toByteArray();
 
-    // BigInteger's toByteArray writes data in two's complement,
-    // so positive values may have a leading 0x00 byte.
-    if (bytes[0] == 0x00) {
+    // BigInteger's toByteArray writes data in two's complement, so positive values may have a
+    // leading 0x00 byte, which we want to strip off. However, we only want to strip these
+    // leading values off if the number of bytes is greater than 1 so that we don't produce an
+    // empty array when the BigInteger value is 0.
+    if (bytes[0] == 0x00 && bytes.length > 1) {
```
sublimator: ooooh, ouch

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
```diff
@@ -0,0 +1,111 @@
...
+        {
+            BigInteger.valueOf(255L),
+            BaseEncoding.base16().decode("FF")
+        },
+        {
+            BigInteger.valueOf(1024L),
+            BaseEncoding.base16().decode("0400")
+        },
+        {
+            BigInteger.valueOf(16385L),
```
sublimator: Why not use hex literals ? 0x4001L etc ?

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
```diff
@@ -0,0 +1,111 @@
...
+  private final BigInteger expectedUint;
+  private final byte[] exptedEncodedBytes;
+
+  private AsnUintCodec codec;
+
+  /**
+   * Construct an instance of this parameterized test with the supplied inputs.
+   *
+   * @param expectedUint       The expected value, as a {@link BigInteger}, of {@code
+   *                           exptedEncodedBytes}, once encoded.
```
sublimator: > exptedEncodedBytes

typo

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
sappenin: Good question - I think the test is clearer if we use Base10 numbers here because the purpose of this test is to ensure that a `BigInteger` will encode into binary. Since `BigInteger` is always operated upon in Base10, I want to be sure that the B10 number properly encodes/decodes. So, the `expectedUint` value is always represented in Base10 on purpose, and we only use the HEX value to construct the byte-array.

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
sappenin: Fixed.

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
sappenin: Actually, looking more closely at this test, @adrianhopebailie, this seems to be treating this like a Uint64, but instead the `AsnUintCodec` should be using a length-prefix, no? (since the number of octets can be arbitrary)

sublimator: I think the serializer classes handle that right ? (checking ... )

sublimator: Hrmmm, this codec doesn't actually seem to be registered at all for the OER rules?

sublimator: re: hex literals, I just thought they'd more easily eyeball match the hex encoding

sublimator: @sappenin 

Yeah, it seems I recalled correctly, and that the serializers handle all the length prefix encoding. 

See:
https://github.com/hyperledger/quilt/blob/b53a0511007c005c1f5455ec7f59639fb935e5b6/codec-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContextFactory.java#L65
https://github.com/hyperledger/quilt/blob/b53a0511007c005c1f5455ec7f59639fb935e5b6/codec-framework/src/main/java/org/interledger/encoding/asn/serializers/oer/AsnOctetStringOerSerializer.java#L88

sappenin: Oh dang, you're totally right. I've been spending so much time in the Codecs I forgot about the Serializers.

Kk - I think this PR is good to go.

PR #167, opened 2018-08-08
Testing/ilpv4 enhancements

### LINKED ISSUE #166: Change number-type-mapping for AsnUint8Codec and AsnUint16Codec
In order to (de)serialize the [CCP](https://github.com/interledgerjs/ilp-protocol-ccp) to/from ASN.1 OER, it is necessary for the Java codecs to support Uint16 datatypes (in order to (d)encode the identifier of a `route-property`).

This ticket proposes to do this by modifying the underlying Java data-type of `AsnUint8Codec` and `AsnUint16Codec` to be `Short` and `Integer` respectively.

In this way, if we want to write Java code (using the current codec framework) that emits an 8-bit number (in a single octet), we would use a `Short`. If we wanted to write Java code that emits a 16-bit number (2 octets), we would use an `Integer`, and so on. The following illustrates the new proposed mapping of ASN.1 OER datatypes to Java number types:

*  8-bit **Fixed** Length Number (1 octet) --> `Short` [**_changed_**]
* 16-bit **Fixed** Length Number (2 octets) --> `Integer` [**_changed_**]
* 32-bit **Fixed** Length Number (4 octet) --> `Long` [_unchanged_]
* 64-bit **Fixed** Length Number (8 octet) --> `BigInteger` (if X <= `2^64-1`) [_unchanged_]
* Var Octets Number --> `BigInteger` (if X > `2^64-1`) [_unchanged_]

sappenin:
This PR primarily fixes #166, but also contains other minor cleanups for ILPv4, including slightly enhanced abstract unit test classes for the Codec classes.


btp-core-codecs/src/main/java/org/interledger/btp/asn/codecs/AsnBtpErrorDataCodec.java
```diff
@@ -43,7 +44,7 @@ public AsnBtpErrorDataCodec(long requestId) {
         requestId,
         new AsnUtf8StringCodec(new AsnSizeConstraint(3)), //Code
         new AsnUtf8StringCodec(AsnSizeConstraint.UNCONSTRAINED), //Name
-        new AsnIA5StringCodec(AsnSizeConstraint.UNCONSTRAINED), //TODO Parse time
+        new AsnIA5StringCodec(AsnSizeConstraint.UNCONSTRAINED), //TriggeredAt
```
sublimator: BTP isn't using the 'new skool' fixed length sizes ?
https://github.com/interledger/rfcs/issues/440#issuecomment-405197100

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
```diff
@@ -26,12 +26,11 @@
    * Construct an instance of this parameterized test with the supplied inputs.
    *
    * @param expectedUint         The expected value, as a {@link BigInteger}, of {@code
-   *                             expectedEncodedBytes}, once encoded.
+   *                             exptedEncodedBytes}, once encoded.
```
sublimator: typo

btp-core-codecs/src/main/java/org/interledger/btp/asn/codecs/AsnBtpErrorDataCodec.java
sublimator: Guess not: https://github.com/interledger/rfcs/blob/master/asn1/BilateralTransferProtocol.asn#L72

adrianhopebailie: There is a lot of stuff in there that is outdated. Many of the BTP message types are never used, including `Prepare`.

https://github.com/interledger/rfcs/issues/445#issuecomment-409651286

codec-framework/src/test/java/org/interledger/encoding/asn/codecs/AsnUintCodecTest.java
sappenin: Fixed.

PR #171, opened 2018-08-19
Fixes #170

### LINKED ISSUE #170: Align AsnUint8Codec with other fixed-number codecs
All of the fixed-byte number codecs extend `AsnOctetStringBasedObjectCodec`, but the recently introduced `AsnUint8Codec` does not. We should adjust `AsnUint8Codec` to conform and also remove the `AsnUint8OerSerializer`.

sappenin:
* Make `AsnUint8Codec` extend the correct interface (`AsnOctetStringBasedObjectCodec `)
* Use `AsnOctetStringOerSerializer` for `AsnUint8Codec`
* Remove AsnUint8OerSerializer


sappenin:
@adrianhopebailie I think this change is correct, but maybe there was a good reason to have `AsnUint8Codec` extend `AsnPrimitiveCodec` instead of using `AsnOctetStringOerSerializer` like the other fixed-number codecs do?

adrianhopebailie:
> maybe there was a good reason to have AsnUint8Codec extend AsnPrimitiveCodec instead of extending AsnOctetStringBasedObjectCodec like the other fixed-number codecs do?

It seemed like an over-complication and the AsnUint8Codec seemed likely to have better performance but I never tested it and this is more consistent so works for me.

Issue #145, opened 2018-05-09
Unit tests for ildcp-core-codecs


adrianhopebailie:
Required for v1

PR #182, opened 2019-03-04
Enhancements to support IL-DCP

### LINKED ISSUE #144: Unit tests for ildcp-core
Required for v1

sappenin:
* Update to 0.16.0-SNAPSHOT
* Fix Jackson security (see [security alerts](https://github.com/hyperledger/quilt/network/alert/jackson-datatypes/pom.xml/com.fasterxml.jackson.core:jackson-databind/open)) 
* Add unit tests for all IL-DCP classes
* Improve Javadoc in various classes
* Remove deprecated InterledgerPacketMapper.
* Fixes #144 


ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequest.java
```diff
@@ -36,11 +38,19 @@ static IldcpRequestBuilder builder() {
...
+  default Instant getExpiresAt() {
+    return Instant.now().plusSeconds(30);
+  }
 
   @Immutable
   abstract class AbstractIldcpRequest implements IldcpRequest {
 
+    @Override
+    @Default
+    public Instant getExpiresAt() {
```
sublimator: Immutable doesn't play nicely with Java 8 default interface methods ?

ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequestPacket.java
```diff
@@ -2,24 +2,29 @@
...
 /**
- * An extension of {@link InterledgerPreparePacket} that can be used as an IL-DCP request over
- * Interledger.
+ * An extension of {@link InterledgerPreparePacket} that can be used as an IL-DCP request over Interledger.
  */
 public interface IldcpRequestPacket extends InterledgerPreparePacket {
 
+  InterledgerAddress PEER_DOT_CONFIG = InterledgerAddress.of(InterledgerAddressPrefix.PEER.with("config").getValue());
+
   InterledgerCondition EXECUTION_CONDITION = InterledgerCondition.of(
```
sublimator: So basically `new byte[32]`

ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequest.java
sappenin: Just maintaining the existing style of the project.

I think the original intent was to define interfaces that had no mention of Immutables, and the define an `abstract` class extension of the interfaces so that future code could create alternative implementations, potentially without needing to depend on the `immutables` library.

That said, as a pattern, it has mostly just created code-duplication. The `immutables` dependency isn't strictly needed at runtime, and having it as a `provided` dependency isn't that big of a deal. I wouldn't be opposed to moving to the "interface-only" style, but it's probably something we should do in another ticket, all at once.

ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequestPacket.java
sappenin: Yep, though I _suppose_ what is currently there is [a bit more explicit](https://github.com/interledger/rfcs/blob/master/0031-dynamic-configuration-protocol/0031-dynamic-configuration-protocol.md#packet) to show spec-conformance, and the hashing is only done once in the JVM so not hurting performance.

ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequest.java
sublimator: Thanks the info

ildcp-core/src/main/java/org/interledger/ildcp/IldcpRequestPacket.java
sublimator: Ah, ok, might be nice to have a named constant then. I personally can't read base64 :)
I see elsewhere comments about what it is:
```
  /**
   * The execution_condition of an ILP packet for IL-DCP is always	   * The execution_condition of an ILP packet for IL-DCP is always
   * <tt>Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=</tt> in Base64 format.	   * <tt>Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=</tt> in Base64 format, which is the SHA-256 hash of a 32-byte
   * array with all 0 values.
   */	   */
```

Of course not really related to this pull request. Just commenting as I was scanning

sappenin:
I'm closing this PR in-favor of #183. It turns out that after writing some unit tests for the Codecs, some other changes are in order, so that PR is more complete, and also fixes #145.

PR #183, opened 2019-03-09
Enhancements to support IL-DCP (new & improved)

### LINKED ISSUE #144: Unit tests for ildcp-core

### LINKED ISSUE #145: Unit tests for ildcp-core-codecs

sappenin:
* Update to 0.16.0-SNAPSHOT
* Fix Jackson security (see [security alerts](https://github.com/hyperledger/quilt/network/alert/jackson-datatypes/pom.xml/com.fasterxml.jackson.core:jackson-databind/open)) 
* Add unit tests for all IL-DCP classes
* Improve Javadoc in various classes
* Remove deprecated InterledgerPacketMapper.
* Fixes #144 
* Fixes #145


ildcp-core-codecs/src/main/java/org/interledger/ildcp/asn/framework/IldcpCodecContextFactory.java
```diff
@@ -0,0 +1,70 @@
...
+   * @param context the context to register the codecs into
+   *
+   * @return The supplied {@code context} with BTP codecs registered into it.
+   */
+  public static CodecContext register(CodecContext context) {
+    // For encoding this into a `data` field.
+    context.register(IldcpResponse.class, AsnIldcpResponseCodec::new);
+
+    context.register(InterledgerPacket.class, AsnIldcpPacketCodec::new);
+    context.register(IldcpRequestPacket.class, AsnIldcpRequestPacketCodec::new);
```
sappenin: We need to think a bit more about this. An `IL-DCP` request is basically just a PreparePacket, and it doesn't play well with the InterledgerCodecContextFactory, which sort of trumps this CodecContext since both objects are of type `InterledgerPreparePacket`.

I'm leaning towards getting rid of IL-DCP packets as a formal concept, and perhaps just consider a way to encode/decode the `IldcpResponse` into the data-portion of a Fulfill packet.

ildcp-core-codecs/src/main/java/org/interledger/ildcp/asn/framework/IldcpCodecContextFactory.java
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * @param context the context to register the codecs into
+   *
+   * @return The supplied {@code context} with BTP codecs registered into it.
+   */
+  public static CodecContext register(CodecContext context) {
+    // For encoding this into a `data` field.
+    context.register(IldcpResponse.class, AsnIldcpResponseCodec::new);
+
+    context.register(InterledgerPacket.class, AsnIldcpPacketCodec::new);
```
sappenin: I think this might be a mistake...I'll look this weekend.

ildcp-core-codecs/src/main/java/org/interledger/ildcp/asn/framework/IldcpCodecContextFactory.java
adrianhopebailie: Should IldcpRequestPacket extend/implement InterledgerPacket?
Maybe not, since the condition etc are hardcoded

sappenin: So, in the most recent update to this PR, I've kept the concept of an `IldcpRequestPacket` that extends `InterledgerPreparePacket` and `IldcpResponsePacket` that extends `InterledgerFulfillPacket`. The Codecs work now with both, and I like the Java interfaces now because it's mostly obvious when you're dealing with which.

Also, the polymorphic handlers (e.g., see [InterledgerResponsePacketHandler.java](https://github.com/hyperledger/quilt/blob/master/ilp-core/src/main/java/org/interledger/core/InterledgerResponsePacketHandler.java)) still work with ILDCP packets _because_ they extend the actual packets classes.

All that to say, let's keep this pattern for now, and as I work it into a Connector/Sender/Receiver, we can always update from there.

sappenin:
@adrianhopebailie @sublimator This one is ready for review now.

PR #207, opened 2019-08-02
Replaces fixed identifier to CanonicalNames and SimpleNames for corresponding classes

sudheesh001:
- Avoids hard coding of strings for canonical names and simple names.

### DIFF
--- a/jackson-datatypes/src/main/java/org/interledger/quilt/jackson/address/InterledgerAddressModule.java
+++ b/jackson-datatypes/src/main/java/org/interledger/quilt/jackson/address/InterledgerAddressModule.java
@@ -31,7 +31,7 @@ import com.fasterxml.jackson.databind.module.SimpleModule;
  */
 public class InterledgerAddressModule extends SimpleModule {
 
-  private static final String NAME = "InterledgerAddressModule";
+  private static final String NAME = InterledgerAddressModule.class.getName();
 
   /**
    * No-args Constructor.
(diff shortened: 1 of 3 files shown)


jackson-datatypes/src/main/java/org/interledger/quilt/jackson/address/InterledgerAddressModule.java
```diff
@@ -31,7 +31,7 @@
  */
 public class InterledgerAddressModule extends SimpleModule {
 
-  private static final String NAME = "InterledgerAddressModule";
+  private static final String NAME = InterledgerAddressModule.class.getSimpleName();
```
sappenin: I'm not sure what the best practice is here, but it looks like the FasterXML projects that aren't deprecated tend to prefer `.class.getName()` instead.

For example, see [JavaTimeModule.java](https://github.com/FasterXML/jackson-modules-java8/blob/master/datetime/src/main/java/com/fasterxml/jackson/datatype/jsr310/JavaTimeModule.java#L113) and [JodaModule.java](https://github.com/FasterXML/jackson-datatype-joda/blob/master/src/main/java/com/fasterxml/jackson/datatype/joda/JodaModule.java#L26).

Any thoughts on using that everywhere?

sudheesh001: Ah the reason I was using `.getSimpleName()` instead of `getName()` was because I thought there was a reason why the string was only a `SimpleName` previously. I'd be happy to change all of them to consistently be `getName()`. Would you want me to go ahead and make that change?

sappenin: > Would you want me to go ahead and make that change?

Yes. I'm not super-clear what's the best path forward here, so let's just go ahead and mirror what FasterXML code does.

sudheesh001: I made the changes in the latest commit. Looking forward to your review @sappenin

PR #208, opened 2019-08-04
Fixes #203 Make @Immutable a part of ilp-core and remove annotation modules

### LINKED ISSUE #203: Move `ilp-annotations` into `ilp-core`.
It's probably overkill to require an entirely new module just for one annotation definition for immutables.

sudheesh001:
- Updates the corresponding imports in other modules using @Immutable
- Adds `ilp-core` as dependency to `codec-framework`
- Removes `ilp-annotations` and refactors `@Immutable` to `ilp-core`


codec-framework/pom.xml
```diff
@@ -51,6 +51,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
sappenin: The codec-framework should not have any overt dependency on `ilp-core`. Technically speaking, `coded-framework` should be able to just stand on its own.

Looking more deeply here, the only reason this dependency is required is because it's used in some of the tests (primarily defined [here](https://github.com/hyperledger/quilt/blob/master/codec-framework/src/test/java/org/interledger/encoding/MyCustomObject.java#L57)).

Instead, we can update that object to just look like this:

```
public interface MyCustomObject {

  static ImmutableMyCustomObject.Builder builder() {
    return ImmutableMyCustomObject.builder();
  }

... // Other class definitions here.

 @Value.Immutable
  abstract class AbstractMyCustomObject implements MyCustomObject {

  }
```

sudheesh001: Perfect. I'll make the changes and you could review them before we could go ahead and close the issue.

sudheesh001:
@sappenin I've made the changes suggested. Looking forward to your review.

PR #209, opened 2019-08-06
Fixes #200 Aligns codec registration methods to use register from oer

### LINKED ISSUE #200: Align all codec registration methods
We should adjust the manner in which ILP Codecs are registered, for consistency, as follows:

- [ ] In `InterledgerCodecContextFactory` add `.register(CodecContext)` and make it `public`.
- [ ] In `BtpCodecContextFactory`, make the `.register(CodecContext)` method `public`.

Finally,  we should also update the contract for `CodecContextFactory` to mirror the above changes. To make that happen, we should attempt the following (I think this will work, but feel free to report back if this causes too many problems):

* Remove the `CodecContextFactory#OCTET_ENCODING_RULES` constant.
* Remove `public static CodecContext getContext(String encodingRules)` and replace it with a `.oer()` and a `.register(CodecContext)` method like the above factories.

Note this means you'll want to maintain the contents of  `CodecContextFactory#getContext(String encodingRules)` but adjust it to work _without_ the String input. E.g., Instead of throwing an exception, this method should only return OER (i.e., remove the if/else on line 75 of `CodecContextFactory`).

Once that is done, you'll want to start using `CodecContextFactory.oer()` in all of the above factories (`InterledgerCodecContextFactory`, `IldcpCodecContextFactory`, and `BtpCodecContextFactory`).

sudheesh001:

### DIFF
--- a/btp-core-codecs/src/main/java/org/interledger/btp/asn/framework/BtpCodecContextFactory.java
+++ b/btp-core-codecs/src/main/java/org/interledger/btp/asn/framework/BtpCodecContextFactory.java
@@ -39,6 +39,7 @@ import org.interledger.encoding.asn.framework.CodecContext;
 import org.interledger.encoding.asn.framework.CodecContextFactory;
 
 import java.time.Instant;
+import java.util.Objects;
 
 /**
  * A factory class for constructing a CodecContext that can read and write Bilateral Transfer Protocl (BTP) 2.0 objects
@@ -54,7 +55,7 @@ public class BtpCodecContextFactory {
    * @return A new instance of {@link CodecContext}.
    */
   public static CodecContext oer() {
-    final CodecContext btpCodecContext = CodecContextFactory.getContext(CodecContextFactory.OCTET_ENCODING_RULES);
+    final CodecContext btpCodecContext = CodecContextFactory.oer();
     return register(btpCodecContext);
   }
 
@@ -65,7 +66,9 @@ public class BtpCodecContextFactory {
    *
    * @return The supplied {@code context} with BTP codecs registered into it.
    */
-  static CodecContext register(CodecContext context) {
+  public static CodecContext register(final CodecContext context) {
+    Objects.requireNonNull(context, "context must not be null");
+
     return context
         .register(BtpError.class, AsnBtpErrorCodec::new)
         .register(BtpMessage.class, AsnBtpMessageCodec::new)
--- a/codec-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContextFactory.java
+++ b/codec-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContextFactory.java
@@ -20,8 +20,6 @@ package org.interledger.encoding.asn.framework;
  * =========================LICENSE_END==================================
  */
 
-import static java.lang.String.format;
-
 import org.interledger.encoding.asn.codecs.AsnCharStringBasedObjectCodec;
 import org.interledger.encoding.asn.codecs.AsnIA5StringBasedObjectCodec;
 import org.interledger.encoding.asn.codecs.AsnIA5StringCodec;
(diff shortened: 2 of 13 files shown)


codec-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContextFactory.java
```diff
@@ -50,50 +48,57 @@
...
-
   /**
-   * Get the {@link CodecContext} loaded with neccessary codecs and serializers for the given
-   * encoding rules.
+   * Create an instance of {@link CodecContext} which uses ASN.1 OER encoding.
    *
-   * @param encodingRules encoding rules to use for serialization
+   * @return A new instance of {@link CodecContext}.
+   */
+  public static CodecContext oer() {
```
sappenin: Thinking a bit more about this specific class, I don't see a need for a generic `register` method because this is a base factory. So, I think the only method we need in this class is this one:

```java
  /**
   * Constructs a {@link CodecContext} that is configured to read and write ASN.1 OER encodings.
   *
   * @return A {@link CodecContext}
   */
  public static CodecContext oer() {
    final AsnObjectCodecRegistry mappings = new AsnObjectCodecRegistry()
        .register(byte[].class, () -> new AsnOctetStringCodec(AsnSizeConstraint.UNCONSTRAINED))
        .register(Short.class, AsnUint8Codec::new)
        .register(Integer.class, AsnUint16Codec::new)
        .register(Long.class, AsnUint32Codec::new)
        .register(BigInteger.class, AsnUint64Codec::new)
        .register(String.class, () -> new AsnUtf8StringCodec(AsnSizeConstraint.UNCONSTRAINED));

    final AsnObjectSerializationContext serializers = new AsnObjectSerializationContext()
        .register(AsnCharStringBasedObjectCodec.class, new AsnCharStringOerSerializer())
        .register(AsnIA5StringCodec.class, new AsnCharStringOerSerializer())
        .register(AsnIA5StringBasedObjectCodec.class, new AsnCharStringOerSerializer())
        .register(AsnOctetStringCodec.class, new AsnOctetStringOerSerializer())
        .register(AsnOpenTypeCodec.class, new AsnOpenTypeOerSerializer())
        .register(AsnOctetStringBasedObjectCodec.class, new AsnOctetStringOerSerializer())
        .register(AsnSequenceCodec.class, new AsnSequenceOerSerializer())
        .register(AsnSequenceOfSequenceCodec.class, new AsnSequenceOfSequenceOerSerializer())
        .register(AsnUint8Codec.class, new AsnOctetStringOerSerializer())
        .register(AsnUint16Codec.class, new AsnOctetStringOerSerializer())
        .register(AsnUint32Codec.class, new AsnOctetStringOerSerializer())
        .register(AsnUint64Codec.class, new AsnOctetStringOerSerializer())
        .register(AsnUtf8StringCodec.class, new AsnCharStringOerSerializer())
        .register(AsnUtf8StringBasedObjectCodec.class, new AsnCharStringOerSerializer());

    return new CodecContext(mappings, serializers);
  }
```

ilp-core-codecs/src/main/java/org/interledger/core/asn/framework/InterledgerCodecContextFactory.java
```diff
@@ -59,7 +59,18 @@
...
+  }
+
+  /**
+   * Register the ILP codecs into the provided context.
+   *
+   * @param context the context to register the codecs into
+   * @return The supplied {@code context} with ILP Codecs registered into it.
+   */
+  public static CodecContext register(CodecContext context) {
+    return context
```
sappenin: This project generally enforces two style rules. First, method params should always be non-null (see [here](https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained) for why), and if `null` is encountered, the  program should halt. 

Second, method parameters should be `final` where possible to help reduce errors (these styles aren't everywhere in this codebase, but should be). So, something like this:

```java
public static CodecContext register(final CodecContext context) {
    Objects.requireNonNull(context, "context must not be null");
   
    ...
}
```

btp-core-codecs/src/main/java/org/interledger/btp/asn/framework/BtpCodecContextFactory.java
```diff
@@ -65,7 +65,7 @@ public static CodecContext oer() {
    *
    * @return The supplied {@code context} with BTP codecs registered into it.
    */
-  static CodecContext register(CodecContext context) {
+  public static CodecContext register(CodecContext context) {
```
sappenin: see comment below about `Objects.requireNonNull`.

sudheesh001:
Thank you for the detailed feedback and clarifications @sappenin. I made the changes and hopefully should be better now. Looking forward to your review.

codec-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContextFactory.java
sappenin: LGTM!

PR #210, opened 2019-08-07
Fixes #205 Remove dev-ops module having maven script utils and unused SecureRandomProvider

### LINKED ISSUE #205: Consider removing `devops` module
It's unclear if the `dev-ops` module is adding any value. Maven-scripting is typically done through alias commands, and the DeterministicSecureRandomProvider doesn't appear to be used in any tests at present.

sudheesh001:


sappenin:
@adrianhopebailie Any opinion around this PR? 

The `dev-ops` module isn't adding any value IMHO, although I'm curious if you think we should keep `DeterministicSecureRandomAlgorithm.java` around somewhere, such as in `ilp-core` or somewhere else. That class is not used in _this_ project, but I can't remember if you were finding it useful somewhere downstream.

sappenin:
One more point of context - this class doesn't appear to be used anywhere in Github except for forks of `ilp-core`: https://github.com/search?q=DeterministicSecureRandomAlgorithm&type=Code

adrianhopebailie:
`DeterministicSecureRandomAlgorithm` was used to test things that use a PRNG. It allows you to specify the "random" bytes that will be generated so you can write unit tests against it.

sappenin:
@adrianhopebailie Was that ever used anywhere though? Maybe when we had crypto-conditions in the project?

dev-ops/util/README.md
```diff
@@ -1,26 +0,0 @@
-Collection of utility scripts wrapping common task (mvn local install, test,...)
-
```
sappenin: I think we need to update the project's README file as well.

sudheesh001:
Looking into the history of the file, it looks like the file was originally in the `ilp-core/src/test/` and later migrated to `dev-ops/src/main`. I don't find any test cases explicitly using this file. I suggest we move it back to the `ilp-core/src/test` in case we'd like to maintain this as a test util for future tests. What do you think @adrianhopebailie @sappenin ?

sappenin:
> I suggest we move it back to the ilp-core/src/test in case we'd like to maintain this as a test util for future tests

I spoke to @adrianhopebailie and we agree - since it's unused, let's remove it. We can always dig it up from the Git history if we need it.

sudheesh001:
Sounds great. I also realized that with this change there was a need to move the `checkstyle.xml` which was previously inside `dev-ops` to the parents to ensure `mvn checkstyle:checkstyle` worked as expected. Updated the PR for the same.

PR #211, opened 2019-08-09
Fixes #199 Adds startsWith(prefix) and tests to the InterledgerAddress interface

### LINKED ISSUE #199: Add `startsWith(prefix)` to InterledgerAddress
The `InterledgerAddress` interface required the following method:

```
 /**
   * <p>Tests if this InterledgerAddress starts with the specified {@code addressPrefix}.</p>
   *
   * @param addressPrefix An {@link InterledgerAddressPrefix} to compare against.
   *
   * @return {@code true} if this InterledgerAddress begins with the specified prefix.
   */
  default boolean startsWith(final InterledgerAddressPrefix addressPrefix) {
    Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
    ...
  }
```

sudheesh001:
- Add `startsWith()` to accept `InterledgerAddressPrefix`
- Add corresponding tests for `InterledgerAddress` that `startsWith(prefix)`

### DIFF
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -141,6 +141,18 @@ public interface InterledgerAddress {
   }
 
   /**
+   * <p>Tests if this InterledgerAddress starts with the specified {@code addressPrefix}.</p>
+   *
+   * @param addressPrefix An {@link InterledgerAddressPrefix} to compare against.
+   *
+   * @return {@code true} if this InterledgerAddress begins with the specified prefix else {@code false}.
+   */
+  default boolean startsWith(final InterledgerAddressPrefix addressPrefix) {
+    Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
+    return this.startsWith(addressPrefix.getValue());
+  }
+
+  /**
    * <p>Return a new {@link InterledgerAddress} by suffixing the supplied {@code addressSegment}
    * onto the current address.</p>
    *
--- a/ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
+++ b/ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
@@ -196,6 +196,47 @@ public class InterledgerAddressTest {
     assertThat(address.startsWith(InterledgerAddress.of("test1.foo.bob")), is(false));
   }
 
+  @Test
+  public void testStartsWithInterledgerAddressPrefix() {
+    final InterledgerAddress address = InterledgerAddress.of("g.foo.bob");
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g")), is(true));
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo")), is(true));
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo.bob")), is(true));
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo.bar")), is(false));
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("test.foo")), is(false));
+
+    final InterledgerAddress smallAddress = InterledgerAddress.of("g.foo");
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g")), is(true));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo")), is(true));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo.bob")), is(false));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo.bar")), is(false));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("test.foo")), is(false));
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testAddressPrefixStartsWithNull() {
+    final InterledgerAddress address = InterledgerAddress.of("g.foo");
+    InterledgerAddressPrefix addressPrefix = null;
+    try {
+      address.startsWith(addressPrefix); // addressPrefix will be null
+    } catch (NullPointerException e) {
+      assertThat(e.getMessage(), is("addressPrefix must not be null!"));
+      throw e;
+    }
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testAddressPrefixStartsWithEmpty() {
+    final InterledgerAddress address = InterledgerAddress.of("g.foo.bob");
+    InterledgerAddressPrefix addressPrefix = InterledgerAddressPrefix.of("");
+    try {
+      address.startsWith(addressPrefix);
+    } catch (IllegalArgumentException e) {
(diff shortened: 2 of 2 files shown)


ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -140,6 +140,18 @@ default boolean startsWith(final InterledgerAddress interledgerAddress) {
...
+  /**
+   * <p>Tests if this InterledgerAddress starts with the specified {@code addressPrefix}.</p>
+   *
+   * @param addressPrefix An {@link InterledgerAddressPrefix} to compare against.
+   *
+   * @return {@code true} if this InterledgerAddress begins with the specified prefix else {@code false}.
+   */
+  default boolean startsWith(final InterledgerAddressPrefix addressPrefix) {
+    Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
+    return this.getPrefix().isPresent() && this.getPrefix().get().startsWith(addressPrefix.getValue());
```
sappenin: This looks good, although let's use a more fluent style, like this:

```java
return this.getPrefix().map(prefix -> prefix.startsWith(addressPrefix.getValue())).orElse(false);
```

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -140,6 +140,18 @@ default boolean startsWith(final InterledgerAddress interledgerAddress) {
     return this.getValue().startsWith(interledgerAddress.getValue());
   }
 
+  /**
+   * <p>Tests if this InterledgerAddress starts with the specified {@code addressPrefix}.</p>
```
sappenin: I think it would be useful to clarify in the Javadoc here the somewhat surprising case of an ILP address `g.foo.bob` and a prefix of `g.foo.bob`, which will evalue `false` here.

I'm actually still on the fence if this is the correct behavior. I _think_ it is, but it's surprising nonetheless. For example, for a minute I thought this was the right implementation of this method:

```java
return this.getValue().startsWith(addressPrefix.getValue());
```

But after thinking about it a bit more, I think the way you did it below is technically correct due to calling `getPrefix`, but it's still surprising which is giving me pause.

sappenin: @adrianhopebailie Do you have any opinion as to what the correct behavior should be here?

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
```diff
@@ -196,6 +196,16 @@ public void testStartsWithInterledgerAddress() {
     assertThat(address.startsWith(InterledgerAddress.of("test1.foo.bob")), is(false));
   }
 
+  @Test
+  public void testStartsWithInterledgerAddressPrefix() {
```
sappenin: We should add coverage for the case where `getPrefix`is very small. For example, an address of just `g.foo`:
```
   address = InterledgerAddress.of("g.foo");
    assertThat(address.startsWith(InterledgerAddressPrefix.of("g")), is(true));
    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo")), is(false));
    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo.bob")), is(false));
    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo.bar")), is(false));
    assertThat(address.startsWith(InterledgerAddressPrefix.of("test.foo")), is(false));
```

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
```diff
@@ -196,6 +196,16 @@ public void testStartsWithInterledgerAddress() {
...
+    final InterledgerAddress address = InterledgerAddress.of("g.foo.bob");
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g")), is(true));
+    assertThat(address.startsWith(InterledgerAddressPrefix.of("g.foo")), is(true));
```
sappenin: I also create #212 for later tracking - I don't think we should implement that in this PR, but might be a good fast-follower once this is solid.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
sappenin: @sudheesh001 after thinking a bit more about this last night, I realized I made an error in my comment above. 

It occurred to me that proposed implementation in this PR of [startsWith](https://github.com/hyperledger/quilt/pull/211/files#diff-e147a50f73f571a8f86bff72f2299dadR150) is incorrect. My rationale is that this method is being called on an Interledger Address, with an input of a `prefix`. In other words, the method is allowing developers to ask, "Does this ILP address start with this prefix?" 

I think the proper answer to this is _not_ to compare the supplied input against the ILP Address's prefix (which is instead a derived property of the ILP address). Instead, the question above is more properly answered by computing whether or not the entire address starts with the prefix. In other words, we should be comparing the input against the entire address itself.

So, I think the proper way to implement this method is actually like this:

```java
return this.getValue().startsWith(addressPrefix.getValue());
``` 

Does that reasoning make sense?

This also should resolve any "surprise" encountered by certain unit tests. For example: `g.foo.com` _does_ technically start with the prefix `g.foo.com`.

sudheesh001: Ah I see. Thinking about it a little more I think this behaviour is consistent with what someone using `startsWith()` would expect. Making the changes and sending in a PR.

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
```diff
@@ -196,6 +196,35 @@ public void testStartsWithInterledgerAddress() {
...
+    final InterledgerAddress smallAddress = InterledgerAddress.of("g.foo");
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g")), is(true));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo")), is(true));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo.bob")), is(false));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("g.foo.bar")), is(false));
+    assertThat(smallAddress.startsWith(InterledgerAddressPrefix.of("test.foo")), is(false));
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void testStartsWithInterledgerAddressPrefixWithNull() {
```
sappenin: I think this test is slightly incorrect. It's actually failing with an NPE inside of `InterledgerAddressPrefix`, which is not what we want to be testing here. 

Instead, this should probably be:

```java
 @Test(expected = NullPointerException.class)
  public void testAddressStartsWithNull() {
    final InterledgerAddress address = InterledgerAddress.of("g.foo");
    InterledgerAddressPrefix addressPrefix = null;
    try {
      address.startsWith(addressPrefix); // addressPrefix will be null
    } catch (NullPointerException e) {
      assertThat(e.getMessage(), is("addressPrefix must not be null!"));
      throw e;
    }
  }
```

sappenin:
One final minor issue in the test, and then I think this is ready to merge. Nice work!

PR #213, opened 2019-08-12
Update for new version

sappenin:

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ Next, add the following Maven dependency:
   <dependency>
     <groupId>org.interledger</groupId>
     <artifactId>java-ilp-core</artifactId>
-    <version>0.13.0-SNAPSHOT</version>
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -62,7 +62,7 @@ Next, add the following Maven dependency:
   <dependency>
     <groupId>org.interledger</groupId>
     <artifactId>java-ilp-core</artifactId>
-    <version>0.13.0-SNAPSHOT</version>
+    <version>0.17.0-SNAPSHOT</version>
```
sudheesh001: This seems like a bump in the versioning from `13`-->`17` while the rest seem to be consistent at `16`-->`17`. Is there a specific reason?

sappenin: This was an omission in previous updates. It should have been `0.16.0-SNAPSHOT` but we overlooked updating it.

PR #216, opened 2019-08-15
Fixes #212 Remove Optionality from InterledgerAddress getPrefix

### LINKED ISSUE #212: Consider making InterledgerAddress.getPrefix non-optional.
Give the `SEGMENTS_UNDERFLOW` logic in the InterledgerAddress preconditions checking, we can likely remove the "optionality" of #getPrefix because there will always be a prefix.

Thus, it's likely that the method contract can instead be this:

```java
default InterledgerAddressPrefix getPrefix() {
    ...
```

sudheesh001:
- The `SEGMENTS_UNDERFLOW` logic ensures that `InterledgerAddress` always contains a prefix.
- Removes optionality from the `getPrefix()` method and updates `hasPrefix()` accordingly.
- Updates corresponding tests and implements explicit checking of `hasPrefix()` and `getValue()` checks.
- Removes `hasPrefix()` since the`InterledgerAddress` always has a prefix.

### DIFF
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -181,24 +181,12 @@ public interface InterledgerAddress {
    *
    * @return An optionally present parent-prefix as an {@link InterledgerAddressPrefix}.
    */
-  default Optional<InterledgerAddressPrefix> getPrefix() {
+  default InterledgerAddressPrefix getPrefix() {
     // An address will always contain at least one period (.), so we can always return its prefix.
     final String value = getValue();
-    return Optional.of(
-        InterledgerAddressPrefix.builder()
+    return InterledgerAddressPrefix.builder()
             .value(value.substring(0, value.lastIndexOf(".")))
-            .build()
-    );
-  }
-
-  /**
-   * <p>Determines if this ILP Address has a parent-prefix.</p>
-   *
-   * @return {@code true} if this address has more than two segments after the allocation scheme. Otherwise return
-   *     {@code false}.
-   */
-  default boolean hasPrefix() {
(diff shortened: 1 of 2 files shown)


ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -198,7 +196,7 @@ default InterledgerAddress with(final String addressSegment) {
    *     {@code false}.
    */
   default boolean hasPrefix() {
```
sappenin: I'm thinking we should just remove this method (`hasPrefix`) entirely. I don't think it's possible for `InterledgerAddress` to be constructed without a prefix, so asking if it has one doesn't seem to be valuable.

Thoughts?

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
```diff
@@ -322,39 +322,41 @@ public void testDestinationAddressWithoutEnoughSegments() {
...
   }
 
   @Test
   public void testGetPrefixFromLongAddress() {
     final InterledgerAddress address = InterledgerAddress.of("g.alpha.beta.charlie.delta.echo");
-    assertThat(address.getPrefix().get().getValue(), is("g.alpha.beta.charlie.delta"));
+    assertThat(address.getPrefix().getValue(), is("g.alpha.beta.charlie.delta"));
   }
 
   @Test
```
sappenin: This should be removed per the above comment.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
sudheesh001: I agree. Made the changes for the same and amended the detailed commit message. 😄

PR #217, opened 2019-08-15
Fixes #196 Implements getInteractions on a given InterledgerAddress

### LINKED ISSUE #196: Add `getInteractions` to `InterledgerAddress`
It should be trivially easy to call `getInteractions` on an ILP Address and get a List or Collection of the segments that come after `~`, if any (also known as [Interactions](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#interactions).

sudheesh001:
- Create a `getInteractions` method for use with `InterledgerAddress`.
- Implement the corresponding test cases for the interactions.

### DIFF
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
@@ -23,6 +23,7 @@ package org.interledger.core;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Lazy;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
@@ -150,6 +151,26 @@ public interface InterledgerAddress {
   default boolean startsWith(final InterledgerAddressPrefix addressPrefix) {
     Objects.requireNonNull(addressPrefix, "addressPrefix must not be null!");
     return this.startsWith(addressPrefix.getValue());
+  }
+
+  /**
+   * Return the Interactions as a list of segments for a given {@link InterledgerAddress}.
+   *
+   * @return List of segments returned as {@code List<String>}.
+   */
+  default List<String> getInteractions() {
+    List<String> interactions = new ArrayList<>();
+    String addressValue = this.getValue();
+
+    if (addressValue.contains("~")) {
+      // @sappenin: Can there be multiple interactions listed as g.a.b~ipr.x.y~ipr1.z.w ?
+      List<String> segments = Arrays.stream(addressValue.split("~")).skip(1).collect(Collectors.toList());
+      for (String segment : segments) {
+        List<String> details = Arrays.asList(segment.split(AbstractInterledgerAddress.SEPARATOR_REGEX));
+        interactions.addAll(details);
+      }
+    }
+    return interactions;
   }
 
   /**
--- a/ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
+++ b/ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
@@ -33,6 +33,8 @@ import org.interledger.core.InterledgerAddress.AllocationScheme;
 
 import org.junit.Test;
 
+import java.util.List;
+
 /**
  * JUnit tests to test {@link InterledgerAddress}.
  */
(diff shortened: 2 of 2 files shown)


sudheesh001:
@sappenin Here is an initial attempt at implementing `getInteractions()`. I left some comments in the patch where I'd appreciate your suggestions.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
```diff
@@ -152,6 +153,26 @@ default boolean startsWith(final InterledgerAddressPrefix addressPrefix) {
...
+   * Return the Interactions as a list of segments for a given {@link InterledgerAddress}.
+   *
+   * @return List of segments returned as {@code List<String>}.
+   */
+  default List<String> getInteractions() {
+    List<String> interactions = new ArrayList<>();
+    String addressValue = this.getValue();
+
+    if (addressValue.contains("~")) {
+      // @sappenin: Can there be multiple interactions listed as g.a.b~ipr.x.y~ipr1.z.w ?
```
sappenin: > Can there be multiple interactions listed as `g.a.b~ipr.x.y~ipr1.z.w`?

My reading of the [grammar defined in RFC-15](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#address-requirements) says, "yes", although I have a question out to the broader community to confirm. See: https://forum.interledger.org/t/questions-about-interactions-in-ilp-addresses/717

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
```diff
@@ -400,6 +402,28 @@ public void testToString() {
     assertThat(InterledgerAddress.of("g.foo").toString(), is("InterledgerAddress{value=g.foo}"));
   }
 
+  @Test
+  public void testGetInteractions() {
+    InterledgerAddress destinationAddress = InterledgerAddress.of("g.us-fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.~ipr.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2");
+    // @sappenin: Is the final match for "~ipr" or "ipr" as the first interaction?
```
sappenin: I added this question to the [forum post](https://forum.interledger.org/t/questions-about-interactions-in-ilp-addresses/717) as well. 

My understanding is that there's nothing normative about interactions - what we read in IL-RFC-15 is just an example, and implementations are free to represent interactions however they want.

To this end, we might just create some "Java ILP Address" rules of our own here. E.g., everything _after_ the first `~` is the interaction string. It can be de-composed by the `.` character.

In that world, the first interaction is `ipr`, but there should be an alternative method that provides access to the entire interaction string, which would be `~ipr.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2`.

**That said**, let's see what comes out of the forum discussion before we move forward with this PR. There's a real possibility (one that I'm leaning towards) that interactions are just something that each consumer of this library can merely create on their own as a higher-order thing (not in this library). If we go that route, then this functionality doesn't belong inside of this class or library at all, and we can maybe just close this PR and #196.

Stay tuned...

sudheesh001: Perfect. Thank you for the detailed response and for creating the forum post. It'd be nice to hear the thoughts of the community on this RFC and have more clarity on the definition of interactions.

ilp-core/src/main/java/org/interledger/core/InterledgerAddress.java
sappenin: I agree with the response in the forum. At this level of the library, I don't think we need to worry about "interactions". Sorry for the false issue in #196.

ilp-core/src/test/java/org/interledger/core/InterledgerAddressTest.java
sappenin: I agree with the response in the forum. At this level of the library, I don't think we need to worry about "interactions". Sorry for the false issue in #196.

sappenin:
I agree with the response in the forum. At this level of the library, I don't think we need to worry about "interactions". Sorry for the false issue in #196 - I'll close that and this PR.

PR #218, opened 2019-08-16
Fixes #214 Adds Overt test cases for of() and removes SEGMENTS_UNDERFLOW in Prefix

### LINKED ISSUE #214: Add test coverage for `InterledgerAddress#of` and `InterledgerAddressPrefix#of`
Assuming this coverage doesn't already exist, on both of these classes, we need something like this: 

```java
 @Test(expected = NullPointerException.class)
  public void testOfWithNull() {
    InterledgerAddressPrefix.of(null);
    fail("should have thrown an exception but did not!");
  }

  @Test(expected = IllegalArgumentException.class)
  public void testOfWithEmptyString() {
    try {
      InterledgerAddressPrefix.of("");
      fail("should have thrown an exception but did not!");
    } catch (IllegalArgumentException e) {
      assertThat(e.getMessage(), is("The '' AllocationScheme is invalid!"));
      throw e;
    }
  }

  @Test(expected = IllegalArgumentException.class)
  public void testOfWithBlankString() {
    try {
      InterledgerAddressPrefix.of(" ");
      fail("should have thrown an exception but did not!");
    } catch (IllegalArgumentException e) {
      assertThat(e.getMessage(), is("The ' ' AllocationScheme is invalid!"));
      throw e;
    }
  }
```

sudheesh001:
- Removes `SEGMENTS_UNDERFLOW` check for `InterledgerAddressPrefix` as this cannot be reached.
- Adds explicit test cases for `InterledgerAddress#of` and `InterledgerAddressPrefix#of`.
- Update error message if a corner case slips through the checks and ask users to report a bug.

### DIFF
--- a/ilp-core/src/main/java/org/interledger/core/InterledgerAddressPrefix.java
+++ b/ilp-core/src/main/java/org/interledger/core/InterledgerAddressPrefix.java
@@ -280,11 +280,6 @@ public interface InterledgerAddressPrefix {
         return String.format(Error.INVALID_SEGMENT.getMessageFormat(), invalidSegment.get());
       }
 
-      // validates the minimum number of segments for a destination address
-      if (segmentsSize < ADDRESS_MIN_SEGMENTS) {
-        return Error.SEGMENTS_UNDERFLOW.getMessageFormat();
-      }
-
       // validates max address length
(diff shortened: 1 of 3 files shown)


sudheesh001:
Great! Thanks for the review. I made the changes and this patch should be ready now.

PR #225, opened 2019-08-23
Fixes #202

### LINKED ISSUE #202: Add handle/fulfill lambdas to PacketMapper and PacketHandler
Currently, creating a new ILP Response Packet handler and/or mapper is a bit wonky, looking like this:

```
new InterledgerResponsePacketHandler() {

            @Override
            protected void handleFulfillPacket(InterledgerFulfillPacket interledgerFulfillPacket) {

            }

            @Override
            protected void handleRejectPacket(InterledgerRejectPacket interledgerRejectPacket) {

            }
          }.handle(responsePacket);
```

Instead, we should augment these classes to instead do this:

```
InterledgerResponsePacket response = ... // do some ILP request.
response.handle((fulfillPacket) -> { 
        logger.info("fulfill");
    }, 
    (rejectPacker) ->     { 
        logger.info("reject");
    }
);
```

and 

```
InterledgerResponsePacket response = ... // do some ILP request.
response.map(fulfillPacket) -> "fulfill", (rejectPacker) -> "reject");
```

sappenin:
* Replace Packet mapper and handlers with methods on InterledgerResponsePacket.
* Replace Packet mapper and handlers with methods on BtpResponsePacket.


sappenin:
I still need to make this adjustment in `IldcpResponsePacketMapper` and `BtpPacketMapper`

ilp-core/src/test/java/org/interledger/core/InterledgerResponsePacketTest.java
```diff
@@ -0,0 +1,75 @@
...
+    assertThat(rejectPacket.getData(), is(EMPTY_DATA));
+  }
+
+  @Test
+  public void handleFulfill() {
+    final AtomicBoolean fulfillCalled = new AtomicBoolean();
+
+    fulfillPacket.handle(
+        (fulfillPacket) -> fulfillCalled.set(true),
+        (rejectPacket) -> fail("Reject handler should be called.")
```
sudheesh001: Shouldn't this be `Reject handler should not be called.`?

ilp-core/src/test/java/org/interledger/core/InterledgerResponsePacketTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    assertThat(fulfillCalled.get(), is(true));
+  }
+
+  @Test
+  public void handleReject() {
+    final AtomicBoolean rejectCalled = new AtomicBoolean();
+
+    rejectPacket.handle(
+        (fulfillPacket) -> fail("Fulfill handler should be called."),
```
sudheesh001: Similar here, Shouldn't this be `Fulfill handler should not be called.`?

ilp-core/src/test/java/org/interledger/core/InterledgerResponsePacketTest.java
sappenin: Nice catch! This is fixed.

ilp-core/src/test/java/org/interledger/core/InterledgerResponsePacketTest.java
sappenin: Nice catch! This is fixed.

sappenin:
@sudheesh001 I think this is ready for final review.

ilp-core/src/test/java/org/interledger/core/InterledgerResponsePacketTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.junit.Test;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Unit tests for {@link InterledgerResponsePacket}.
+ */
+public class InterledgerResponsePacketTest {
+
+  private static final byte[] EMPTY_DATA = new byte[64];
```
sudheesh001: Curious: Any specific reason why this is a `64` byte array instead of the usual `32`?

sappenin: No specific reason. Could be data/size, really. Probably a better name would be just `DATA` since "empty data" usually looks like `new byte[]`.

Issue #229, opened 2019-09-02
Make StreamClient Multithreaded


sappenin:
Currently, SimpleStreamClient is not multi-threaded because it always uses StreamId of `1`. However, in a process (e.g., Artemis) sending money in multiple streams to the same receiver, this would either not work or corrupt the stream (likely, only one process would succeed).

sappenin:
This is complete and merged to `master`.

PR #234, opened 2019-09-13
StreamClientTest now spins up an instance of interledger-rs in a docker container (via testcontainers)

nhartner:
Necessary ILP accounts are created as part of test setup

Signed-off-by: nhartner <nhartner@gmail.com>


nhartner:
It doesn't make sense to me to put a test in the Java connector project for a class in the Quilt project. If there's a bug in StreamClient, it should be caught as part of its project build. Especially when you have an open issue like https://github.com/hyperledger/quilt/issues/229 that will need additional tests to verify the fix.

My reason for making this change the way I did was:
- The build was failing on this test
- The client should be tested against something that is close as possible to a real server
- The test can download and spin up the Interledger-rs docker image quickly (under 5s)
- The Java ILP connector is not dockerized yet and I assumed it still does not fully implement streams
- The change set is small and self-contained

But let's pick the best method to test this class/module. I'll defer to your judgement if you think we are betting off testing StreamClient by hitting StreamReceiver.

sappenin:
@nhartner Your rationale makes sense to me - let's keep going with this one using Interledger-rs.

Maybe also:

1. Once Java Connector get dockerized, let's consider using that instead of, or in addition to, the Rust Connector in this harness to provide solid integration testing of the client (so we know the client works against the Java stack too).

1. Decide how best to test the StreamServer. I'll move that over to Slack.

PR #244, opened 2019-09-13
Add STREAM core primitives (STREAM-PR1)

sappenin:
* Bump SNAPSHOT versions to 0.18.0-SNAPSHOT
* Add minor updates to ILP-core to support STREAM
* Reorg poms for STREAM
* Add stream-core module
* Defer ConnectionManager to #241 

Signed-off-by: David Fuelling <sappenin@gmail.com>

Issue #242, opened 2019-09-13
Ensure that StreamPacket's frame data doesn't exceed ILP data packet size, which is 32k.


sappenin:
See ASN.1 OER definition of the ILPv4 packet's data field here: https://github.com/interledger/rfcs/blob/master/asn1/InterledgerProtocol.asn#L30

One way to guard against this is just the Immutables check in the InterledgerPacket.java. 

However, we might consider adding logic in the stream-core code possibly, although since this is a guard at the ILPv4 packet, we probably should just leave it there.

sappenin:
Also, see test coverage in `AsnStreamDataFrameCodecTest`

sappenin:
Thinking more about this, maybe we just write some test coverage in `SenderReceiverTest` to exercise this use-case and make sure we like the error messaging. I think that most connectors will reject ILPv4 packets with more than 32k of data, so it's unclear to me if we need to do anything special here, but I think it depends on what the errors look like coming from the STREAM sender, etc.

theotherian:
Upon further discussion this probably isn't a case that we can handle in Quilt very effectively as the packet size is an implementation detail of a connector and should really only be reinforced by the server and not the client. If the client were to enforce this limit as a precondition it could create a scenario in the future where, if the server were to increase the packet size it could handle beyond 32kb, there could be incompatibility forwarding a payload due to a client restricting the size where the server doesn't actually care.

PR #252, opened 2019-09-13
Add STREAM Codecs (STREAM-PR2)

sappenin:
* Rename `codec-framework` to `codecs-framework`
* Add `codecs-stream` to the project.


sappenin:
This PR depends on #244. That should be merged to `master` first, and then we can change the target of this PR to `master`.

codecs-parent/codecs-stream/src/test/java/org/interledger/codecs/stream/frame/AsnStreamDataFrameCodecTest.java
```diff
@@ -0,0 +1,168 @@
...
+                 .data((byte) 1)
+                 .build()
+         },
+
+         // bigger data
+         {
+             StreamDataFrame.builder()
+                 .streamId(UnsignedLong.ONE)
+                 .offset(UnsignedLong.MAX_VALUE)
+                 .data((byte) 1, (byte) 2, (byte) 3, (byte) 4, (byte) 5, (byte) 6, (byte) 7, (byte) 8, (byte) 9,
```
sappenin: this is just copy/paste, but thinking there's a better way to generate random data of a certain size. Will change this.

sappenin: This is fixed. Also updated https://github.com/hyperledger/quilt/issues/242 with a note to revisit this test once that issue is fixed.

sudheesh001: `new Random(System.currentTimeMillis()).nextBytes(byteArray);` would do the trick right? or maybe use `ThreadLocalRandom`?

PR #253, opened 2019-09-14
Add Link modules (STREAM-PR3)

sappenin:
* Add link-core contains all interfaces and abstractions for links.
* Add link-ilp-over-http contains an implementation of ILP-over-HTTP using OkHttp3.
* Unit tests.
* Fix Javadoc in various places.


sappenin:
This PR depends on #244 and #252. Those should be merged to `master` first, and then we can change the target of this PR.

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/auth/JwtHs256BearerTokenSupplier.java
```diff
@@ -0,0 +1,105 @@
...
+    ilpOverHttpAuthTokens = CacheBuilder.newBuilder()
+        // There should only ever be 1 or 2 tokens in-memory for a given client instance.
+        .maximumSize(3)
+        // Expire after this duration, which will correspond to the last incoming request from the peer.
+        .expireAfterAccess(this.outgoingLinkSettings.tokenExpiry().orElse(Duration.of(30, ChronoUnit.MINUTES)))
+        .removalListener((RemovalListener<String, String>) notification ->
+            logger.debug("Removing IlpOverHttp AuthToken from Cache for Principal: {}", notification.getKey())
+        )
+        .build(new CacheLoader<String, String>() {
+          public String load(final String authSubject) {
```
sudheesh001: Maybe you might want to make this as

```java
public String load(@NotNull final String authSubject) {
```

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpLink.java
```diff
@@ -0,0 +1,298 @@
...
+      okHttpRequest = this.constructSendPacketRequest(preparePacket);
+    } catch (IOException e) {
+      throw new LinkException(e.getMessage(), e, getLinkId());
+    }
+
+    try (Response response = okHttpClient.newCall(okHttpRequest).execute()) {
+      if (response.isSuccessful()) {
+        //////////
+        // Success
+        return ilpCodecContext.read(InterledgerResponsePacket.class, response.body().byteStream());
```
sudheesh001: Do we want to test the validity of the `response.body()` object here?

```java
return ilpCodecContext.read(InterledgerResponsePacket.class, Objects.requireNonNull(response.body()).byteStream());
```

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpLink.java
```diff
@@ -0,0 +1,298 @@
...
+      } else {
+        //////////
+        // Reject!
+        final String errorResponseBody = CharStreams.toString(response.body().charStream());
```
sudheesh001: Do we want to do a similar `requireNonNull()` check for the `charStream()`?

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpLinkFactory.java
```diff
@@ -0,0 +1,106 @@
...
+    final IlpOverHttpLink ilpOverHttpLink = new IlpOverHttpLink(
+        operatorAddressSupplier,
+        ModifiableIlpOverHttpLinkSettings.create().from(ilpOverHttpLinkSettings), // Modifiable for testing
+        okHttpClient,
+        objectMapper,
+        ilpCodecContext,
+        bearerTokenSupplier
+    );
+
+    return ilpOverHttpLink;
```
sudheesh001: Maybe we could directly `return new IlpOverHttpLink()`? Is there a specific reason for this approach? Do we envision this to be changed and used for some kind of validation or checks before returning?

link-parent/link-ilp-over-http/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+      <artifactId>ilp-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>link-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>codecs-framework</artifactId>
+    </dependency>
```
sudheesh001: Nit: Do we want to move `codecs-framework` as the first dependency to maintain lexicographic order?

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/auth/JwtHs256BearerTokenSupplier.java
sappenin: Assuming I'm understanding your comment, I consider `@NotNull` annotations in this project to be redundant because we treat null as a bug. Read more here: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained

link-parent/link-core/src/main/java/org/interledger/link/Link.java
```diff
@@ -0,0 +1,108 @@
...
+ * <p>An abstraction for communicating with a remote Interledger peer in a connection-state-less manner, meaning links
+ * do not hold connection state during their lifecyle.</p>
+ *
+ * <p>If you want to make a {@link Link} that can holds connection state (such as for a link protocol using
+ * Websockets), consider {@link StatefulLink} instead.</p>
+ */
+public interface Link<L extends LinkSettings> extends LinkSender {
+
+  /**
+   * <p>A unique identifier for this {@link Link}. This value is generally set only once, (e.g., in a Connector to
```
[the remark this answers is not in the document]
sappenin: Good points. Let's change to this: 

```suggestion
   * <p>A unique identifier for this {@link Link}.
```

link-parent/link-core/src/main/java/org/interledger/link/LinkFactoryProvider.java
```diff
@@ -0,0 +1,42 @@
...
+        .orElseThrow(() -> new LinkException(
+            String.format("No registered LinkFactory linkType=%s", linkType), LinkId.of("n/a"))
+        );
+  }
+
+  public LinkFactory registerLinkFactory(final LinkType linkType, final LinkFactory linkFactory) {
+    Objects.requireNonNull(linkType, "linkType must not be null");
+    Objects.requireNonNull(linkFactory, "linkFactory must not be null");
+
+    return this.linkFactories.put(linkType, linkFactory);
```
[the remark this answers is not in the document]
sappenin: Simple answer is I'm not sure. In the Connector implementation at least (which is the only think using this) factories are only registered at startup time and then never change.

That said, we should probably leave this alone just in case some other use case comes along....or until we see this causing problems.

link-parent/link-core/src/test/java/org/interledger/link/LinkFactoryProviderTest.java
```diff
@@ -0,0 +1,65 @@
...
+  @Test
+  public void registerAndGetLinkFactory() {
+    linkFactoryProvider.registerLinkFactory(TEST_LINK_TYPE, linkFactoryMock);
+    assertThat(linkFactoryProvider.getLinkFactory(TEST_LINK_TYPE)).isEqualTo(linkFactoryMock);
+  }
+
+  @Test(expected = LinkException.class)
+  public void getUnregisteredLinkFactory() {
+    try {
+      linkFactoryProvider.getLinkFactory(TEST_LINK_TYPE);
```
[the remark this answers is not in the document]
sappenin: Either way - I like the `ExceptedException` pattern.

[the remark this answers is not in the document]
theotherian: #263

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpLinkFactory.java
sappenin: No specific reason - it's likely that at some point during development I might have been doing something before line 98, but _after_ linke 89. As-is though, we can certainly change this.

link-parent/link-core/src/test/java/org/interledger/link/LinkFactoryProviderTest.java
```diff
@@ -0,0 +1,65 @@
...
+    } catch (NullPointerException e) {
+      assertThat(e.getMessage()).isEqualTo("linkType must not be null");
+      throw e;
+    }
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void registerLinkFactoryWithNullLinkFactory() {
+    try {
+      linkFactoryProvider.registerLinkFactory(TEST_LINK_TYPE, null);
```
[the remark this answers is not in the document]
theotherian: #263

link-parent/link-core/src/test/java/org/interledger/link/LoopbackLinkTest.java
```diff
@@ -0,0 +1,149 @@
...
+        () -> Optional.of(OPERATOR_ADDRESS),
+        LinkSettings.builder().linkType(LoopbackLink.LINK_TYPE).build(),
+        packetRejector
+    );
+  }
+
+  @Test(expected = RuntimeException.class)
+  public void registerLinkHandler() {
+    try {
+      link.registerLinkHandler(incomingPreparePacket -> null);
```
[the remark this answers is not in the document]
theotherian: #263

PR #255, opened 2019-09-16
Add STREAM Client (STREAM-PR4)

sappenin:
* Add STREAM client module
* Add (currently ignored) SimpleStreamClient Integration test.
* Unit tests for congestion controller


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
```diff
@@ -0,0 +1,309 @@
...
+    // A "window" is just an amount of time where some value is in-flight, and a StreamClient wants to put more
+    // in-flight. Thus, the amount left in a window will trend towards zero as the amountInFlight increases.
+    final UnsignedLong amountLeftInWindow = maxInFlight.get().minus(amountInFlight.get());
+
+    // Synchronization is _probably not_ needed here since we're not assigning any values that could get corrupted, and
+    // usage of this return value should be considered as a snapshot at a given moment of time anyway...
+    return this.maxPacketAmount
+        // If maxInFlight is specified, take the min of `amountLeftInWindow` and `maxInFlight`.
+        .map(maxPacketAmount -> StreamUtils.min(amountLeftInWindow, maxPacketAmount))
+        .orElse(amountLeftInWindow);
```
theotherian: Whenever I see
> Synchronization is _probably not_ needed here
I immediately read the code probably 10 times.

I don't understand enough about STREAM yet to really speak intelligently, but the answer I have so far is you're _probably_ right. As far as I can tell, the worst case would be that, for some amount of time, you're allowing a larger amount than you actually should, but I'd assume that it's within reason that this isn't expecting to be immediately consistent on those changes? Otherwise it seems like you could create a severe bottleneck trying to block.

All of that aside, the comment says 
> // If maxInFlight is specified, take the min of `amountLeftInWindow` and `maxInFlight`. 

but we're comparing `StreamUtils.min(amountLeftInWindow, maxPacketAmount)`, so one of those is incorrect.

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -0,0 +1,508 @@
...
+        // Create the ILP Prepare packet
+        final byte[] streamPacketData = this.toEncrypted(sharedSecret, streamPacket);
+        final InterledgerCondition executionCondition;
+        executionCondition = generatedFulfillableFulfillment(sharedSecret, streamPacketData).getCondition();
+
+        final InterledgerPreparePacket preparePacket = InterledgerPreparePacket.builder()
+            .destination(destinationAddress)
+            .amount(amountToSend.bigIntegerValue())
+            .executionCondition(executionCondition)
+            .expiresAt(Instant.now().plusSeconds(30L))
```
theotherian: Should this be configurable?

PR #262, opened 2019-09-18
multithreaded stream sender tests

nhartner:
Update StreamSenderIT to spin up an interledger node to test against
Add test to verify sender can handle multithreaded sends 
Now that IT can pass, enable IT via failsafe plugin
Fix some Immutables and Jackson config so that Jackson can handle InterledgerAddress type


stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderIT.java
```diff
@@ -168,7 +231,4 @@ public void sendMoney() {
 //
 //    logger.info("Payment Sent: {}", sendMoneyResult);
   // }
-
-  // TODO: Send too much
```
sappenin: Can we make tickets for these?

PR #257, opened 2019-09-17
Add stream receiver (STREAM-PR-5)

sappenin:
* Add Stream Reciever plus unit test coverage.

PR #265, opened 2019-09-19
Add stream receiver (STREAM-PR-5) #257

theotherian:
Not sure how I managed to close the last one 😕


stream-parent/stream-receiver/src/main/java/org/interledger/stream/server/SpspStreamConnectionGenerator.java
```diff
@@ -0,0 +1,112 @@
...
+
+    // Bytes 0 through 17
+    final byte[] randomBytes = Arrays.copyOf(localPartBytes, 18);
+    final byte[] sharedSecret = Hashing.hmacSha256(secretGenerator(serverSecretSupplier)).hashBytes(randomBytes)
+        .asBytes();
+    // Bytes 18 through 31
+    final byte[] authTag = Arrays.copyOfRange(localPartBytes, 18, localPartBytes.length);
+
+    // The Address without the final 18 bytes.
+    String addressWithoutBytes = receiverAddressAsString.substring(0, receiverAddressAsString.length() - 19);
```
sublimator: Do all these `final` really add much :) ? 
I have been programming in TypeScript/JavaScript mostly lately, and yeah, I habitually use const where possible. The alternative is `var` or `let`, where you have to use *something* to declare a reference, whereas with Java you *can* eschew it.  

Does Java not have a tool to add them in automatically where they are "effectively final" ? Seems like a pain. 

Context being that seems `addressWithoutBytes` seems to be lacking `final`.

sappenin: > Do all these final really add much :) ?

Not really. It's _mostly_ a style thing. E.g., here's a [pretty informative SO](https://stackoverflow.com/questions/8354412/do-java-finals-help-the-compiler-create-more-efficient-bytecode) about why `final` is mostly not useful, but not harmful. 

I personally like `final` modifiers, but at this point in my career I consider it a benign habit that's just part of my style. Mostly I like them because it helps me not have to think about potential simple bugs, like  accidentally mutating a reference somewhere "below" its declaration.

Also, I like them as modifiers to methods, especially constructors, in order to avoid mistakes in assignment. E.g., if you forget to use `this` in the constructor assignment example below, this is a silly bug that the compiler will catch:

```
class Foo {
  private bar;
  Foo(final int bar){
    bar = bar;
  }
}
```

However, the compiler won't catch this mistake if the code is structured like this:

```
class Foo {
  private bar;
  Foo(int bar){
    bar = bar;
  }
}
```

There of plenty of people I highly respect, however, that will offer counter-arguments like, "but modern IDEs will warn on this condition" or "the JIT will actually ignore your modifiers if it makes the code more optimal" or "This makes the code more difficult to read" or whatever.

Like I said, this is purely a style thing, so using `final` or not using it makes _almost_ no difference in practice, especially if code is well tested.

sublimator:
Thanks for the response :)
I dig your style :)

PR #278, opened 2019-09-21
Moar test coverage

theotherian:


ildcp-core/src/test/java/org/interledger/ildcp/InterledgerShampooPacket.java
```diff
@@ -0,0 +1,18 @@
+package org.interledger.ildcp;
+
+import org.interledger.core.Immutable;
+import org.interledger.core.InterledgerResponsePacket;
+
+@Immutable
+public interface InterledgerShampooPacket extends InterledgerResponsePacket {
```
sappenin: Let’s add some Javadoc to explain what this is?

theotherian: Added; it's only in the `test` package for testing fall through cases on `isAssignableFrom` checks.

theotherian:
Closing PR to try namespacing trick so that the branch doesn't go into protected mode

PR #282, opened 2019-09-23
javadoc packet size limit

theotherian:
Addresses #242 


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/StreamSender.java
```diff
@@ -19,6 +19,10 @@
   /**
    * Send "money" (i.e., some unit of value) from a source ILP address to a destination address.
    *
+   * Note that, per https://github.com/hyperledger/quilt/issues/242, as of the publication of this client, connectors
```
sappenin: This is pretty good but feels a bit disconnected from the constructor. What if we tweaked this a bit to indicate why this might be a problem, like this perhaps:

```
   * Note that, per https://github.com/hyperledger/quilt/issues/242, as of the publication of this client, connectors will reject ILP packets that exceed 32kb. This implementation does not overtly check to restrict the size of the `data` field in any particular {@link InterledgerPreparePacket}, for two reasons. First, this implementation never packs a sufficient number of STREAM frames into a single Prepare packet for this 32kb limit to be an issue; Second, if the ILPv4 RFC ever changes to increase this size limitation, we don't want sender/receiver software to have to be updated across the Interledger.
```

Also, let's move this to the Javadoc of the Constructor.

(ps - feel free to tweak my language above if you see fit).

Issue #261, opened 2019-09-18
Ensure StreamSender enforces Exchange Rates


sappenin:
See comments here: https://github.com/interledger-rs/interledger-rs/issues/273

Specifically, this [TODO](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-stream/src/client.rs#L121) in the Rust code (which will have a corresponding entry int he Java code once committed to master). 

Reference [this portion](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md#34-exchange-rates) of IL-RFC-29 for Exchange Rates.

sappenin:
The fix on the Client for this task is more formally defined in #291

sappenin:
See #291

sappenin:
@theotherian there is a new comment in the associated RUST issue from Kincaid. Let’s look at our implementation to confirm we’re accounting for this use-case.

PR #285, opened 2019-09-24
add connection asset details to check exchange rate

theotherian:
I legitimately don't know if this will actually accomplish what's intended in #261, but it does accomplish what was said in https://github.com/interledger-rs/interledger-rs/issues/273 and Kincaid is way smarter than I am at this.


stream-parent/stream-receiver/src/main/java/org/interledger/stream/receiver/StatelessStreamReceiver.java
```diff
@@ -14,12 +14,7 @@
 import org.interledger.stream.StreamPacket;
 import org.interledger.stream.StreamUtils;
 import org.interledger.stream.crypto.StreamEncryptionService;
-import org.interledger.stream.frames.ConnectionCloseFrame;
-import org.interledger.stream.frames.ErrorCode;
-import org.interledger.stream.frames.StreamFrame;
-import org.interledger.stream.frames.StreamFrameType;
-import org.interledger.stream.frames.StreamMoneyFrame;
-import org.interledger.stream.frames.StreamMoneyMaxFrame;
+import org.interledger.stream.frames.*;
```
nhartner: Checkstyle will complain about the star import. I disabled star imports in IDEA to avoid this.

stream-parent/stream-receiver/src/test/java/org/interledger/stream/receiver/StatelessStreamReceiverTest.java
```diff
@@ -22,8 +22,7 @@
 import org.interledger.stream.StreamUtils;
 import org.interledger.stream.crypto.JavaxStreamEncryptionService;
 import org.interledger.stream.crypto.StreamEncryptionService;
-import org.interledger.stream.frames.StreamFrameType;
-import org.interledger.stream.frames.StreamMoneyFrame;
+import org.interledger.stream.frames.*;
```
nhartner: star import

nhartner:
I'm confused why this is necessary. The STREAM packet contains a prepareAmount and StatelessStreamReceiver#receiveMoney already checks that the amount received is >= the prepareAmount and either fulfills or rejects accordingly. If an intermediary connector tried to steal money, wouldn't they be thwarted by the receiver rejecting the packet?

theotherian:
@sappenin I'm sure you would give a more coherent answer to that question than I could.

stream-parent/stream-receiver/src/main/java/org/interledger/stream/receiver/StatelessStreamReceiver.java
```diff
@@ -76,7 +76,8 @@ public StreamConnectionDetails setupStream(final InterledgerAddress receiverAddr
    */
   @Override
   public InterledgerResponsePacket receiveMoney(
-      final InterledgerPreparePacket preparePacket, final InterledgerAddress receiverAddress
+      final InterledgerPreparePacket preparePacket, final InterledgerAddress receiverAddress, final String assetCode,
```
sappenin: I'm pretty sure the whole "asset scale" thing is confusing for people first encountering it and Interledger, so any objection to making an object here to hold the `assetCode` and `assetScale`?

That way we can define the term "scale" in one spot and potentially re-use the object in other places that need these two fields.

Not sure _exactly_ what to call this thing, but maybe `Denomination`...(I care less about the name and more about being able to provide the Javadoc that explains how scale works -- apologies @theotherian because I know you hate long Javadoc...but it's loosely adapted from [here](https://java-connector.ilpv4.dev/concepts/terminology)).

E.g., 

```
/**
*/
@Immutable
/**
   */
  String getAssetCode();

   /**
   *
   *
   *
   * specified above.</p>
   *
   * scale is 0.</p>
   *
   *
   */
  short getAssetScale();
}
```

sappenin:
> I'm confused why this is necessary. The STREAM packet contains a prepareAmount and StatelessStreamReceiver#receiveMoney already checks that the amount received is >= the prepareAmount and either fulfills or rejects accordingly. If an intermediary connector tried to steal money, wouldn't they be thwarted by the receiver rejecting the packet?

Per our conversation today, the StreamSender first needs to get the receiver's ConnectionAssetDetailsFrame in order to compute the right amount to put into the StreamPacket. Until that happens, the sender can't put anything meaningful into the StreamPacket's amount field, which means [this check](https://github.com/hyperledger/quilt/blob/master/stream-parent/stream-receiver/src/main/java/org/interledger/stream/receiver/StatelessStreamReceiver.java#L136) won't be doing anything until that happens.

Issue #286, opened 2019-09-25
Fix Bugs in `SimpleStreamSender`


sappenin:
The current implementation of SimpleStream Client has a few problems:

1. **Payment Timeout**. The `sendMoney` call should be overridden with a sibling variant that accepts a timeout. If not specified, the sender should never timeout. However, if specified, the sender should timeout **[FIXED BY #290]**

2. **Refactor SendMoneyAggregator**: **[FIXED BY #290]**

3. **Exchange Rates**: **[FIXED BY #291]**

4. **While-loop condition**: **[FIXED BY #290]** [This while-loop condition](https://github.com/hyperledger/quilt/blob/master/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java#L234) is technically correct, but is confusing. Instead, it should probably look like this:

```java
// Used to count the total number of STREAM packets send for this payment
final AtomicInteger totalStreamPackets = new AtomicInteger(0);
// Used to track the duration of this STREAM payment.
final Instant start = Instant.now();
while (true) {
    if (amountLeftToSend <= 0)) {
          break;
        }
   //... Keep sending packets.
}
```

4b. TODO: add a Preconditon that `amountLeftToSend` is NEVER < 0, and "throw" if this ever happens. **[FIXED BY #293]**

5. **Sleeping Thread**: **[FIXED BY #290]** If the `amountToSend` is `ZERO`, the sender sleeps for 100ms and then continues the loop. However, we should consider just breaking out here because this is what occurs now, but in a less obvious way. That said, we should tighten the while-loop to not stop until `deliveredAmount` is actually the value we want. In other words, we should keep looping until either `deliveredAmount` is enough, or the timeout is triggered. We need to think carefully though about not overpaying.

6. **ILPv4 Packet Failures**: **[FIXED BY #289]** We need to consider what should happen in the STREAM client when certain ILPv4 errors are encountered. Right now, only `T04` and `F08` are handled, with all other errors simply being retried. This is probably fine, but we should look through the [JS Rejection logic](https://github.com/interledgerjs/ilp-protocol-stream/blob/master/src/connection.ts#L1440) and the [Rust Rejection logic](https://github.com/interledger-rs/interledger-rs/blob/master/crates/interledger-stream/src/congestion.rs#L116) to be sure we're doing this correctly.

sappenin:
**Java ILP Call Notes**

[...]

PR #290, opened 2019-09-27
Is/parallelism around congestion

theotherian:
Addresses some of #286

### DIFF
--- a/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
+++ b/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
@@ -290,6 +290,11 @@ public class AimdCongestionController implements CongestionController {
     return maxPacketAmount;
   }
 
+  @Override
+  public boolean hasInFlight() {
+    return this.amountInFlight.get().compareTo(UnsignedLong.ZERO) > 0;
+  }
+
   public void setMaxPacketAmount(final UnsignedLong maxPacketAmount) {
     this.maxPacketAmount = Optional.of(maxPacketAmount);
   }
--- a/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
+++ b/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
@@ -235,7 +235,6 @@ public class AimdCongestionController implements CongestionController {
         // maxAmount: Maximum amount (inclusive, denominated in same units as the receivedAmount) the connector
         // will forward
         // Equation: new_max_packet_amount = prepare_amount * details.max_amount() / details.amount_received();
-        // TODO: See alternatives in https://github.com/interledger-rs/interledger-rs/issues/49
         final BigDecimal newMaxPacketAmountAsBigDecimal =
             prepareAmountAsBigDecimal.multiply(detailsMaxAmount).divide(detailsAmountReceived, RoundingMode.FLOOR);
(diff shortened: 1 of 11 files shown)


sappenin:
I've created https://github.com/hyperledger/quilt/issues/302 to improve the unit test coverage in `SimpleStreamSender`. That appears to be the primary regression for Codecov, so assuming the rest of these changes look good, then I vote we merge this and improve coverage as a fast-follower. We have _pretty good_ coverage of the client between [SimpleStreamSenderTest](https://github.com/hyperledger/quilt/blob/master/stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderTests.java) and [StreamSenderIT](https://github.com/hyperledger/quilt/blob/master/stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderIT.java).

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
```diff
@@ -290,6 +289,11 @@ public void setCongestionState(final CongestionState congestionState) {
     return maxPacketAmount;
   }
 
+  @Override
+  public boolean hasInFlight() {
```
sappenin: We need a unit test for this one. @theotherian I seem to recall you wrote this - maybe it didn't get pushed?

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -308,46 +419,75 @@ public SendMoneyAggregator(
...
-          return responsePacket;
-        });
+        try {
+          // don't submit new tasks if the timeout was reached within this iteration of the while loop
+          if (canBeScheduled(timeoutReached.get(), streamPacket.sequence())) {
+            // call this before spinning off a task. failing to do so can create a condition in the
+            // soliderOn check where we will incorrectly evaluate hasInFlight on the congestion
+            // controller to not reflect what we've actually scheduled to run, resulting in the loop
+            // breaking prematurely
+            congestionController.prepare(amountToSend);
```
sappenin: We should think a bit more about this. While we want the `congestionController.prepare(amountToSend)` to be called outside the execution, I think we need to account for two cases:

* An error is thrown while calling `executorService.submit(()` (line `430`).
* An Exception is thrown and processed in line `442`.

In either case, it seems like we would want to unwind the call to `congestionController.prepare(amountToSend)` to avoid deadlock per that comment. However, it seems like the call to prepare in the congestion controller should be more tightly connected here.

Any thoughts?

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/AimdCongestionController.java
sappenin: Nm - I see the test.

sappenin:
Let's discuss replacing this PR with #305, which started from these changes, subsumed #303, and then introduced a ConnectionManager to make the threading boundaries clearer.

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
theotherian: One thought is that the call to unwind a prepare be idempotent somehow, but that would require some kind of means of tracking a call to prepare by returning a uuid or something similar. Due to the fact that we're scheduling we can't rely on a try/finally approach since the finally executes outside of the thread if you're trying to guard against an exception on submit.

sappenin:
Closing in-favor of #305

PR #303, opened 2019-09-28
Misc Updates

### LINKED ISSUE #288: Introduce `Denomination` class in ilp-core
```java
/**
*/
@Immutable
/**
   */
  String getAssetCode();

   /**
   *
   *
   *
   * specified above.</p>
   *
   * scale is 0.</p>
   *
   *
   */
  short getAssetScale();
}
```

sappenin:
Misc Updates

* Formatting, Checkstyle, and Javadoc improvements
* Introduce PacketLoss to SimulatedIlpv4Network
* Enhance SenderRecieverTest to simulate various network loss scenarios.
* Introduce SharedSecret into Stream code
* Enhance StreamSender to properly honor timeouts
* Enhance SteramSender to properly prevent multiple sendMoney calls per Connection/SharedSecret
* Add failed Link ITs to SimpleStreamSenderIT
* Introduce Denomination class (Fixed #288)


sappenin:
Closing in-favor of #305

Issue #291, opened 2019-09-27
Implement ExchangeRate Mechanism for StreamSender


sappenin:
We talked about doing the following to support ExchangeRates in the `StreamSender`:

**Problem**
Currently, the sender doesn't account for exchange rates. While this is currently being addressed as part of #261, we need to update the sender to do the following:

* Open the STREAM connection with a StreamPacket containing the following: A 0 amount, a ConnectionNewAddressFrame, and a StreamMoneyFrame (this is happening, but should maybe be pulled out into its own function called openMoneyStream.
* Once the connection asset details are received, the STREAM can then start using the CongestionController to send N number of packets in parallel with a particular packet amount.
* Need MinimumAcceptableAmountGuard interface with a "Always Return 1" basic implementation.
* `boolean isAcceptable(BigInteger sourceAmount, Denomination sourceDenomination, BigInteger reportedAmountDelivered, Denomination reportedAmountDenomination, maxSlippage);
* Add overloaded sendMoney that accepts an instance MinimumAcceptableAmountGuard. Other method has no MinimumAcceptableAmountGuard and calls the other sendMoney with a default implementation that always true.

**However**, before we implement this design, let's understand what the [Javascript implementation is doing here](https://github.com/interledgerjs/ilp-protocol-stream/blob/master/src/connection.ts#L1160).

sappenin:
We need to make the StreamSender aware of its Link Asset type and scale (see `Denomination`).

theotherian:
This ticket as a bunch of moving parts to it that ended up being identified but need to be documented (and have their details sorted out) for an eventual implementation.

- **The MinimumAcceptableAmountGuard interface needs methods to indicate if it accepts specific asset codes.** These would need to be checked before attempting to apply a guard to a transaction, otherwise the result could be wildly inaccurate. 
- **As of right now, it seems the most appropriate place to add the guard is `StreamSender.sendMoney`.** The reason for this is that constraining the guard to the constructor of an implementation of `StreamSender` creates a burden on preemptively passing a guard capable of servicing all calls to `sendMoney` for the lifetime of the sender.
- **The `SimpleStreamSender` needs to parse the `ConnectionAssetDetails` frame returned from the receiver (all `StreamSender` implementations have to).** Without capturing this information, we have no idea whether or not a guard even works properly for a given `sendMoney` invocation.
[...]

PR #305, opened 2019-09-30
Misc improvements to STREAM sender + receiver

### LINKED ISSUE #279: Enhance `SenderReceiverTest` with a lossy-network simulation (e.g., 50% packet rejection)

### LINKED ISSUE #288: Introduce `Denomination` class in ilp-core

sappenin:
**Misc Updates**

* Introduce ConnectionManager into STREAM sender
* Introduce PacketLoss to `SimulatedIlpv4Network` (Fixes #279)
* Introduce Denomination class (Fixes #288)
* Enhance StreamSender to properly honor timeouts and Connection sequences per-JVM
* Enhance SenderRecieverTest to simulate various network loss scenarios.
* Add failed Link ITs to SimpleStreamSenderIT
* Formatting, Checkstyle, and Javadoc improvements


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -192,70 +293,114 @@ default int totalPackets() {
...
+              .sendMoneyDuration(sendMoneyDuration.get())
+              .build()
+          );
+    }
+
+    private void sendMoneyPacketized() {
+      final AtomicBoolean timeoutReached = new AtomicBoolean(false);
 
+      timeout.ifPresent($ -> {
+        ScheduledExecutorService timeoutMonitor = Executors.newSingleThreadScheduledExecutor();
```
nhartner: I wonder if this causes a memory leak creating a ScheduledExecutorService but never shutting it down.

nhartner: Confirmed this will actually lead to a memory leak. To avoid this, we should change the scheduled lambda from `timeoutReached.set(true)` to  `{ timeoutReached.set(true); timeoutMonitor.shutdown(); }`

link-parent/link-core/src/main/java/org/interledger/link/exceptions/LinkException.java
```diff
@@ -85,4 +85,17 @@ public LinkException(
...
+  /**
+   * Overridden to emit the Link Id.
+   *
+   * @return A {@link String} representation of this Exception.
+   */
+  @Override
+  public String toString() {
+    String s = getClass().getName();
+    String message = getLocalizedMessage() + " LinkId=" + getLinkId();
+    return (message != null) ? (s + ": " + message) : s;
```
nhartner: I can't see now message would ever be null

stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SendMoneyAggregatorTest.java
```diff
@@ -52,37 +69,138 @@ public void setUp() {
     when(streamEncryptionServiceMock.encrypt(any(), any())).thenReturn(new byte[32]);
     when(streamEncryptionServiceMock.decrypt(any(), any())).thenReturn(new byte[32]);
     when(linkMock.sendPacket(any())).thenReturn(mock(InterledgerRejectPacket.class));
-
+    final StreamConnection streamConnection = new StreamConnection(StreamConnectionId.of("foo"));
```
nhartner: Intellij shows this as unused. Did you mean to use this or the mock?

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/StreamConnectionManager.java
```diff
@@ -0,0 +1,75 @@
...
+   * <p>One concurrency-related note is that there is a small moment of time after this method has been called, but
+   * where it has not yet processed the close request. During this time, an in-use Connection will process calls to
+   * {@link StreamConnection#nextSequence()} so long as the sequence is not too large. Ultimately it is the
+   * responsibility of {@link StreamConnection} to manage it effective Connection state.</p>
+   *
+   * @param streamConnectionId
+   *
+   * @return
+   */
+  public Optional<StreamConnection> closeConnection(final StreamConnectionId streamConnectionId) {
```
nhartner: Intellij says this method is never called. Seems like it should be.

stream-parent/stream-core/src/main/java/org/interledger/stream/StreamConnection.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * <p>The session established between two endpoints that uses a single shared secret and multiplexes multiple streams
+ * of money and/or data.</p>
+ *
+ * <p>Any given Connection in a JVM manages a unique sequence id that can be incremented for each Stream Packet sent
+ * over the connection.</p>
+ */
+public class StreamConnection {
```
nhartner: Might consider making this implement `AutoCloseable` so that in can be used in a try-with-resources block.

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -91,95 +117,170 @@ public SimpleStreamSender(
...
+      final SharedSecret sharedSecret,
       final InterledgerAddress sourceAddress,
       final InterledgerAddress destinationAddress,
       final UnsignedLong amount
   ) {
     Objects.requireNonNull(destinationAddress);
     Objects.requireNonNull(amount);
 
-    final SendMoneyAggregator sendMoneyAggregator = new SendMoneyAggregator(
+    final StreamConnection streamConnection = this.streamConnectionManager.openConnection(
```
nhartner: Per my comment above, `streamConnection` is never closed.

link-parent/link-core/src/main/java/org/interledger/link/exceptions/LinkException.java
sappenin: Fixed here: https://github.com/hyperledger/quilt/commit/9e83dd74ad8526f0e28bea2e6a04ac880c8c5443

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
sappenin: hmmm...I think it is closed down on [line 363](https://github.com/hyperledger/quilt/pull/305/files#diff-2ec7bd39fe6f2a1ff349260aeddbdccaR363), no?

sappenin: Oh, my bad - the actual object. nm.

sappenin: For `v1.0`, @nhartner and I decided that the `SimpleStreamSender` will not close the Connection nor the Stream since the only receiver is a stateless receiver. We'll revisit this as part of #291

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/StreamConnectionManager.java
sappenin: See my comment above. We'll revisit this as part of #291

stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SendMoneyAggregatorTest.java
sappenin: Fixed.

stream-parent/stream-core/src/main/java/org/interledger/stream/StreamConnection.java
sappenin: Agree. Tracking in #310.

sappenin:
> We should fix the minor memory leak (with ScheduledExecutor) and make sure StreamConnection#closeConnection is called in SimpleStreamSender

This is now addressed. Please re-review and approve if things look OK. 

Last night I fixed the flaky tests, but this branch is way too big, so so if there are any further substantive changes let's track them in a ticket and work on them as a fast-follower.

PR #318, opened 2019-10-03
Is/exchange rates the sequel

theotherian:
Addresses https://github.com/hyperledger/quilt/issues/291


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -1,7 +1,7 @@
 package org.interledger.stream.sender;
 
-import static org.interledger.core.InterledgerErrorCode.F08_AMOUNT_TOO_LARGE_CODE;
 import static org.interledger.core.InterledgerErrorCode.T04_INSUFFICIENT_LIQUIDITY_CODE;
+import static org.interledger.core.InterledgerErrorCode.F08_AMOUNT_TOO_LARGE_CODE;
 import static org.interledger.stream.StreamUtils.generatedFulfillableFulfillment;
```
sudheesh001: There might be an ordering issue here of the imports.

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -289,22 +296,20 @@ default int totalPackets() {
     private AtomicInteger numFulfilledPackets;
     private AtomicInteger numRejectedPackets;
 
+    private Optional<Denomination> receiverDenomination = Optional.empty();
```
sudheesh001: To keep the consistency, wouldn't it be nicer to initialize the variable to `Optional.empty()` in the constructor for `SendMoneyAggregator` instead of having it here.

theotherian:
Also addresses #306 (as well as #291 )

PR #320, opened 2019-10-04
Df/fixes 201

### LINKED ISSUE #201: Remove BigInteger from InterledgerPreparePacket and replace with long or UnsignedLong
In order for this to work, we need to consider one of these three options:

1. Leave the codec framework alone, but create new Codecs for unsigned primitives (e.g., `AsnUint64CodecUL` and `AsnUint32CodecUL`) that map to `UnsignedLong` (instead of a `BigInteger`) and `UnsignedInt` (instead of a `Long`).
1. Refactor the Codec framework to support explicit correlation for properties in a given class (e.g., one class might decode to a long, and another might decode to a BigInteger).
1. Rewrite the Codec system per #164 

If we choose choice #1 (the least amount of work), we should also consider if we should _only_ replace `AsnUint64Codec`, in which case we would have the following:

`AsnUint8Codec` -> `Short`
`AsnUint16Codec` -> `Integer`
`AsnUint32Codec` -> `Long`
`AsnUint64CodecUL` -> UnsignedLong

...or if instead we should do this:

`AsnUint8Codec` -> `byte[]` (limit of 8, converted to int for usage later)
`AsnUint16Codec` -> `Short` (treated as unsigned)
`AsnUint32Codec` -> `Integer`(treated as unsigned)
`AsnUint64Codec` -> `Long`(treated as unsigned)

### More Context
* https://www.nayuki.io/page/unsigned-int-considered-harmful-for-java
* Between Stream, BTP, and ILP, and CCP, all integer types are used (`UInt64`, UInt32`, and `UInt16`, and `UInt8`).

sappenin:
Fixes #201

* Use UnsignedLong instead of BigInteger in packets
* Fix Codecs to support UnsignedLong


codecs-parent/codecs-framework/src/test/java/org/interledger/encoding/asn/serializers/oer/UnsignedLongOerSerializerTest.java
```diff
@@ -0,0 +1,152 @@
...
+                new byte[] {(byte) 0x80, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01}},
+            // 33
+            {UnsignedLong.valueOf("18446744073709551614"),
+                new byte[] {(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF,
+                    (byte) 0xFF, (byte) 0xFF, (byte) 0xFE}},
+            // 34
+            {UnsignedLong.valueOf("18446744073709551615"),
+                new byte[] {(byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF, (byte) 0xFF,
+                    (byte) 0xFF, (byte) 0xFF, (byte) 0xFF}}
+        }
```
sudheesh001: It'd be nice to have an explanation or a link pointing to the documentation or the standard that defines this behavior. Other languages that might view quilt as a baseline and implement their equivalents might need to know exactly how to handle these overflows which are greater than the standard supported bounds.

sappenin: While these _are_ overflows for a typical Java `long` number (which is signed), these aren't overflows for `UnsignedLong`, which can hold up to `2^64 - 1`

PR #330, opened 2019-10-10
Preconditions to prevent over sending

theotherian:


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/FixedSenderAmountPaymentTracker.java
```diff
@@ -90,8 +90,12 @@ public PrepareAmounts getSendPacketAmounts(UnsignedLong congestionLimit,
   }
 
   @Override
-  public void auth(PrepareAmounts prepareAmounts) {
+  public boolean auth(PrepareAmounts prepareAmounts) {
```
sappenin: Can we add a unit test here?

sappenin: nm - maybe this is covered?

Issue #352, opened 2019-10-17
JDK > 8 builds break when generating javadoc throwing MavenReportException


sudheesh001:
Temporary workaround is to disable generation of the javadocs: `-Dmaven.javadoc.skip=true` along with the `mvn install` command.

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.0:jar (attach-javadocs) on project ilp-core: MavenReportException: Error while generating Javadoc:
[ERROR] Exit code: 1 - javadoc: error - An exception occurred while building a component: TagInfo
[ERROR] 	(java.lang.NullPointerException)
[ERROR] Please file a bug against the javadoc tool via the Java bug reporting page
[ERROR] (http://bugreport.java.com) after checking the Bug Database (http://bugs.java.com)
[ERROR] for duplicates. Include error messages and the following diagnostic in your report. Thank you.
[ERROR] java.lang.NullPointerException
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.toolkit.util.Utils.getEnclosingTypeElement(Utils.java:2574)
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.toolkit.util.CommentHelper.getReferencedClass(CommentHelper.java:371)
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.HtmlDocletWriter.seeTagToContent(HtmlDocletWriter.java:1510)
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.HtmlDocletWriter$2.visitLink(HtmlDocletWriter.java:1949)
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.HtmlDocletWriter$2.visitLink(HtmlDocletWriter.java:1797)
[ERROR] 	at jdk.compiler/com.sun.tools.javac.tree.DCTree$DCLink.accept(DCTree.java:500)
[ERROR] 	at jdk.compiler/com.sun.source.util.SimpleDocTreeVisitor.visit(SimpleDocTreeVisitor.java:79)
[ERROR] 	at jdk.javadoc/jdk.javadoc.internal.doclets.formats.html.HtmlDocletWriter.commentTagsToContent(HtmlDocletWriter.java:2017)
[...]

PR #353, opened 2019-10-17
fixes for JDK9+

nhartner:
Javadoc doesn't like @link a generic (i.e. {@link T}) because it can't be linked to an actual type. Replace with {@code T}
Java 9 changed time precision to microseconds but ILP stuff seems to use millisecond precision. For now, revert back to millisecond precision via a common DateUtils.now() method.

address #352


codecs-parent/codecs-framework/src/main/java/org/interledger/encoding/asn/framework/CodecContext.java
```diff
@@ -97,7 +97,7 @@ public CodecContext(AsnObjectCodecRegistry mappings, AsnObjectSerializationConte
 
   /**
    * Deserialize an object of the given type from the given stream and use the appropriate ASN.1
-   * codec to decode the deserialized ASN.1 object into an object of type {@link T}.
+   * codec to decode the deserialized ASN.1 object into an object of type }.
```
sappenin: Missing something here?

nhartner: Good catch.  Find/replace mess up. Fixed.

PR #359, opened 2019-10-23
Add CI Build for Java 14 and updates tests to not use spy

sudheesh001:
- Test failures in java14 are due to mockito incompatibility.
- Replace mockito spy usages with the interface instances.

Signed-off-by: Sudheesh Singanamalla <sudheesh@cs.washington.edu>


sappenin:
@sudheesh001 Can you go into circle and click "rebuild" for the `build` and `build_java10` builds? I think these are just flake-failures.

PR #377, opened 2019-11-21
Implements stream fixtures to be verified as available in the RFCs

sudheesh001:
The rationale behind this testing method is to have the following ability:
1. There is a local fixtures file which will be available in the quilt codebase for testing
2. We want to ensure that the changes to the fixtures file in the RFC repository forces us to update our fixtures file.

This patch introduces `FixtureManager` which performs the integrity `shasum` checks on the fixtures provided in the RFCs and compares it with the local fixtures file used for testing. The test will break with an `Exception` if either the fixtures in the RFC repository changes or if it changes locally within quilt.

### DIFF
--- a/codecs-parent/codecs-stream/src/test/java/org/interledger/codecs/stream/StreamPacketFixturesTest.java
+++ b/codecs-parent/codecs-stream/src/test/java/org/interledger/codecs/stream/StreamPacketFixturesTest.java
@@ -21,6 +21,10 @@ package org.interledger.codecs.stream;
  */
 
 import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
+import static org.interledger.codecs.stream.frame.helpers.FixtureManager.checkInternetConnectivity;
+import static org.interledger.codecs.stream.frame.helpers.FixtureManager.checksum;
+import static org.interledger.codecs.stream.frame.helpers.FixtureManager.downloadFixtureFile;
+import static org.interledger.codecs.stream.frame.helpers.FixtureManager.readProperties;
 
 import org.interledger.codecs.stream.frame.helpers.StreamFrameFixture;
 import org.interledger.codecs.stream.frame.helpers.StreamTestFixture;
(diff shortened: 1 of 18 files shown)


sudheesh001:
This is an initial take on resolving the issue. Suggestions appreciated.

codecs-parent/codecs-stream/src/test/java/org/interledger/codecs/stream/StreamPacketFixturesTest.java
```diff
@@ -83,6 +93,89 @@ public StreamPacketFixturesTest(StreamTestFixture streamTestFixture) {
...
+   * are executed. As a part of the class rule, we expect the validation of the checksum to pass on both the local
+   * test fixtures file and the remote test fixtures file in the RFCs repository. Any change to either the local or
+   * the remote file will result in a failure of this test.
+   */
+  @ClassRule
+  public static ExternalResource externalResource = new ExternalResource() {
+    @Override
+    protected void before() throws Throwable {
+      boolean rfcStatus = checkFixtureRFCStalenessState();
+      boolean localFixtureStatus = checkLocalFixtureFileState();
```
sappenin: Overall, I like the design/intention of this PR. Only thing is that the test should pass if the system running the test is offline or otherwise unable to download the remote fixture. So, what if we just return `false` from `checkFixtureRFCStalenessState()` if it can't download?

sudheesh001:
I have a feeling that it's better to avoid checking the staleness of the fixtures if there is no Internet connection available and go ahead and run the tests with the local fixtures. I pushed an updated commit with this change. Please let me know what you think 😄

codecs-parent/codecs-stream/src/test/java/org/interledger/codecs/stream/StreamPacketFixturesTest.java
sappenin: This has been fixed.

PR #382, opened 2019-11-27
ignore token issuer and audience

theotherian:

### DIFF
--- a/link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IncomingLinkSettings.java
+++ b/link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IncomingLinkSettings.java
@@ -5,7 +5,10 @@ import static org.interledger.link.http.IlpOverHttpLinkSettings.DOT;
 import static org.interledger.link.http.IlpOverHttpLinkSettings.ILP_OVER_HTTP;
 import static org.interledger.link.http.IlpOverHttpLinkSettings.INCOMING;
 import static org.interledger.link.http.IlpOverHttpLinkSettings.SHARED_SECRET;
+import static org.interledger.link.http.IlpOverHttpLinkSettings.TOKEN_AUDIENCE;
+import static org.interledger.link.http.IlpOverHttpLinkSettings.TOKEN_ISSUER;
 
+import okhttp3.HttpUrl;
 import org.immutables.value.Value;
 import org.immutables.value.Value.Modifiable;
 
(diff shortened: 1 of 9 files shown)


link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IncomingLinkSettings.java
```diff
@@ -75,38 +75,47 @@
                   .map(Object::toString)
                   .map(String::toUpperCase)
                   .map(IlpOverHttpLinkSettings.AuthType::valueOf)
-                  .ifPresent(builder::authType);
+                  .ifPresent((authType) -> {
```
sappenin: Minor nit, but the standard `continuation-indent` is 4-spaces for Quilt. This is a bit different from the Connector, but we should keep with the existing style for now.

link-parent/link-ilp-over-http/src/test/java/org/interledger/link/http/IncomingIlpOverIlpOverHttpLinkSettingsSettingsTest.java
```diff
@@ -28,9 +28,23 @@ public void applyCustomSettingsWithFlatDottedNotation() {
...
+  /**
+   * Tests the builder when customAttributes is a flat collection of key/value pairs using dotted-notation,
+   * ignoring properties not applicable to SIMPLE auth
+   */
+  @Test
+  public void applyCustomSettingsWithFlatDottedNotationSimpleAuth() {
+    final Map<String, Object> customSettings = this.customSettingsFlat(IlpOverHttpLinkSettings.AuthType.SIMPLE);
+    final IncomingLinkSettings incomingLinksettings = IncomingLinkSettings.fromCustomSettings(customSettings).build();
+
+    assertThat(incomingLinksettings.authType()).isEqualTo(IlpOverHttpLinkSettings.AuthType.SIMPLE);
```
sappenin: This should probably assert that `tokenIssuer` and `tokenAudience` are absent.

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IncomingLinkSettings.java
sappenin: Fixed via 678b93bdb900cd4c22cb8b14ea7d71e4ef876ccd

link-parent/link-ilp-over-http/src/test/java/org/interledger/link/http/IncomingIlpOverIlpOverHttpLinkSettingsSettingsTest.java
sappenin: This is fixed via de33c1f70bd001a8c17d9030af8e61cd7a921661

link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IncomingLinkSettings.java
```diff
@@ -75,38 +75,47 @@
...
+                    builder.authType(authType);
+
+                    if (authType != IlpOverHttpLinkSettings.AuthType.SIMPLE) {
```
sappenin: It's not in the ticket, but we should probably do this change for the outgoing links too.

Issue #370, opened 2019-11-07
SimpleStreamSender ignores 4XX errors and retries (and fails) repeatedly


nhartner:
When SimpleStreamSender.sendMoney attempts to send a payment and gets a 4XX error (e.g. 401 for bad password), it repeatedly retries sending packets until it eventually times out. Instead, it should abort since 4XX are not retryable.

sappenin:
From Java Community call: 

* `4xx`: Abort immediately. Return a`SendMoneyResult` with 0 packets and a boolean because it's possible we get a 4xx error in the middle of a stream. Consider adding the error message for debugging. 
* For `5xx` errors there should be a retry limit in addition to the exponential back-off. Need to consider not making the `numRetries` lower than what would make the exponential backoff moot. No matter what `numRetries` is we should always just use exponential backoff, even if it's unused.

sappenin:
This ticket handles 4xx HTTP codes. See https://github.com/hyperledger/quilt/issues/289 for the remaining work to address 5xx error codes (maybe).

PR #394, opened 2019-12-05
retry logic galore #370

theotherian:


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/DefaultBackoffController.java
```diff
@@ -0,0 +1,109 @@
...
+      return response;
+    }
+
+    public int rejectsReceived() {
+      return rejects.get();
+    }
+  }
+
+  @Override
+  public InterledgerResponsePacket sendWithBackoff(final Link link, final InterledgerPreparePacket preparePacket,
```
nhartner: I think that the preparePacket.expiresAt would need to be updated/extended before every retry attempt. When SimpleStreamSender creates the preparePacket it sets the expiresAt to 30s in the future. If you configured back such that it retries after 1s, 2s, 4s, 8s, 16s then the logic attempts retry a preparePacket that is already expired.

theotherian: Good catch; totally spaced that behavior. Thanks!

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/DefaultBackoffController.java
```diff
@@ -0,0 +1,109 @@
...
+import io.github.resilience4j.retry.Retry;
+import io.github.resilience4j.retry.RetryConfig;
+
+import java.time.Duration;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+
+public class DefaultBackoffController implements BackoffController {
```
sappenin: If we keep this PR, let's add some Javadoc to outline what the default backoff behavior is?

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/DefaultBackoffController.java
sappenin: I was expecting that if a packet fails, we would consider it a reject, and the "retry" would just be a new packet. Is the design here to do something different?

PR #397, opened 2019-12-05
Fixes rust auth handler for ILP-over-HTTP

sappenin:

### DIFF
--- a/spsp-parent/spsp-core/src/main/java/org/interledger/spsp/PaymentPointerResolver.java
+++ b/spsp-parent/spsp-core/src/main/java/org/interledger/spsp/PaymentPointerResolver.java
@@ -1,5 +1,12 @@
 package org.interledger.spsp;
 
+import okhttp3.HttpUrl;
+
+import java.util.Objects;
+
+/**
+ * Defines how to resolve a payment pointer.
+ */
 public interface PaymentPointerResolver {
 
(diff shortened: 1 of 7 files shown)


spsp-parent/spsp-core/src/main/java/org/interledger/spsp/PaymentPointerResolver.java
```diff
@@ -1,23 +1,56 @@
...
+   * Resolves PaymentPointer to it's endpoint URL.
+   *
    * @param paymentPointer ilp payment pointer
+   *
    * @return endpoint url
+   *
+   * @see "https://interledger.org/rfcs/0026-payment-pointers/#payment-setup-protocol-receiver-endpoint-resolution"
    */
-  String resolve(PaymentPointer paymentPointer);
+  HttpUrl resolveHttpUrl(PaymentPointer paymentPointer);
```
nhartner: Why HttpUrl instead of just URL? Seems like this is a spsp-client implementation detail (that we are using okhttp3 in the DefaultSpspClient) that's leaking into spsp-core. Not a big deal but worth considering if this is ok.

sappenin: After discussing, let's keep HttpUrl. It's also in `SpspClient` and other places.

PR #398, opened 2019-12-06
Is/370 only fail on 4xx

theotherian:


link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpLink.java
```diff
@@ -132,12 +132,11 @@ public InterledgerResponsePacket sendPacket(final InterledgerPreparePacket prepa
...
-        if (response.code() == 401 || response.code() == 403) {
-          // If this code is returned, we know the Link is misconfigured, so throw a LinkException Exception.
-          throw new LinkException(String.format(
-              "Unable to connect to remote ILP-over-HTTP Link: Invalid Bearer Token. response=%s", response
-          ), this.getLinkId());
-        } else if (response.code() >= 400 && response.code() < 500) {
+        if (response.code() >= 400 && response.code() < 500) {
+          if (response.code() == 401 || response.code() == 403) {
+            // If this code is returned, we know the Link is misconfigured
+            logger.error("Unable to connect to remote ILP-over-HTTP Link: Invalid Bearer Token. response={}", response);
```
sappenin: I think it would be helpful to put this exception message into the packet's `message` field so that the packet-sender (who won't have access to the log-files containing the emission on 138) can easily see what went wrong.

link-parent/link-ilp-over-http/src/test/java/org/interledger/link/http/IlpOverHttpLinkTest.java
```diff
@@ -92,23 +92,17 @@ public void sendFailsOnBadWrite() throws Exception {
...
   public void rejectOnNotAuthed() throws Exception {
     mockCall(401);
-    expectedException.expect(LinkException.class);
-    expectedException.expectMessage("Unable to connect to remote ILP-over-HTTP Link: "
-        + "Invalid Bearer Token. response=Response{"
-        + "protocol=h2, code=401, message=stop asking me to set stuff, url=https://existentialcrisis.com/"
-        + "}");
-    link.sendPacket(packet);
+    InterledgerResponsePacket responsePacket = link.sendPacket(packet);
+    assertThat(responsePacket).extracting("code", "message")
```
sappenin: Let's assert the `message` value of the response packet, too.

Issue #386, opened 2019-11-27
Remove space from IlpOverHttpConstants.BEARER


sappenin:
The constant `IlpOverHttpConstants.BEARER` should not have a space in it. The only usage is in IlpOverHttpLink, which should insert a space near line 244.

Sjshekhar:
I will fix it.

sappenin:
Hi @Sjshekhar that sounds great! Please submit a PR at your convenience.

PR #401, opened 2019-12-13
removed space

Sjshekhar:
This pull request is in response to issue #386


link-parent/link-ilp-over-http/src/main/java/org/interledger/link/http/IlpOverHttpConstants.java
```diff
@@ -12,6 +12,6 @@
   String OCTET_STREAM_STRING = OCTET_STREAM.toString();
   MediaType APPLICATION_OCTET_STREAM = MediaType.parse(OCTET_STREAM_STRING);
 
-  String BEARER = "Bearer ";
+  String BEARER = "Bearer";
```
sappenin: You'll need to search through the code-base for usage of `BEARER` and add a space in there. This is why the build is breaking.

PR #416, opened 2020-01-17
Add support for in-memory SPSP Receiver (fixes #415)

### LINKED ISSUE #415: Create SPSPReceiverLink

sappenin:
Add support for an in-memory Stateless Stream Receiver Link. This enables a Connector to fulfill ILP packets without having to forward to an actual SPSP Receiver server.

### DIFF
--- a/link-parent/link-stateless-spsp-receiver/src/main/java/org/interledger/link/spsp/StatelessSpspReceiverLinkSettings.java
+++ b/link-parent/link-stateless-spsp-receiver/src/main/java/org/interledger/link/spsp/StatelessSpspReceiverLinkSettings.java
@@ -1,0 +1,106 @@
+package org.interledger.link.spsp;
+
+import org.interledger.link.LinkId;
+import org.interledger.link.LinkSettings;
+import org.interledger.link.LinkType;
+import org.interledger.link.exceptions.LinkException;
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Derived;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An extension of {@link LinkSettings} for ILP-over-HTTP links.
+ */
+public interface StatelessSpspReceiverLinkSettings extends LinkSettings {
+
+  String ASSET_CODE = "assetCode";
+  String ASSET_SCALE = "assetScale";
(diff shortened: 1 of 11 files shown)


link-parent/link-stateless-spsp-receiver/src/main/java/org/interledger/link/spsp/StatelessSpspReceiverLinkSettings.java
```diff
@@ -0,0 +1,106 @@
...
+
+import org.immutables.value.Value;
+import org.immutables.value.Value.Derived;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * An extension of {@link LinkSettings} for ILP-over-HTTP links.
```
nhartner: This javadoc does not match the class signature. But it raises an interesting question. How will we handle the expected use case for an account that is IlpOverHttpLink on the incoming side but StatelessSpspReceiverLink on the outgoing side?

sappenin: Docs fixed.

sappenin:
> it raises an interesting question. How will we handle the expected use case for an account that is IlpOverHttpLink on the incoming side but StatelessSpspReceiverLink on the outgoing side?

This is a good point, and makes me think Quilt (i.e., a Link) is not the correct place to put this type of behavior. 

One way to do this is to enable the Connector to short-circuit Links with an SPSP Receiver if the destination address matches on some pre-configured SPSP address (e.g., if the Connector is configured to use `test.jc.money` as its SPSP address prefix, then `test.jc.money.david` would get intercepted and treated as an SPSP sub-account. The account used would be `david`, so the balance updates would still work properly. Likewise, "David" could send packets using IlpOverHttp, but receiving would always be handled by an SPSP receiver as opposed to an Outgoing Link.

One nice fallout of this is that if David wanted to connect to the Connector using IL-DCP, then he would get an address of `test.jc.david`, and if David wanted to run a connector and affect that same balance, he could and could run his own SPSP server if he wanted -- his packets would require a destination address of something like `test.jc.david.money.xyz123`

Issue #406, opened 2019-12-18
Make soldierOn Run-loop's "sleep time" configurable


sappenin:
At a minimum, this value should be configurable via an env property, e.g., 500ms.

Note that per @theotherian choosing a longer sleep time probably means we should have longer payment timeout because a longer sleep time increases the chances of hitting the upper time-bound of the payment (i.e., the payment timeout)

sappenin:
Per more recent conversation, we should consider (instead or in addition) just reducing this value to 5ms or something very small. The use-case is when 20 packets need to be sent out, they will be sent sequentially every 100ms, which means that ramp-up will take 20 * 100ms, or 2s.

nkramer44:
Since `SimpleStreamSender` and `SendMoneyAggregator` are not Spring managed beans, configuring this via an application.yml file or another env file gets a bit ugly.  

Would it make more sense to configure this in the `SendMoneyAggregator` constructor (by passing the value through `SimpleStreamSender.sendMoney()`? Any STREAM sender can then configure the value however they want.

PR #420, opened 2020-01-27
Make soldierOn loop sleep configurable

nkramer44:
In `SimpleStreamSender`, we currently sleep for 100 ms in between packet send tries while ramping up.  This should be configurable per #406 

To fix this, I added an `Optional<UnsignedLong>` field to `SimpleStreamSender` and `SendMoneyAggregator` to configure this value, which gets defaulted to 100ms.  Any STREAM sender can configure this value via an env property to instantiate a `SimpleStreamSender`


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -120,33 +146,52 @@ public SimpleStreamSender(final StreamEncryptionService streamEncryptionService,
...
+   *                                end-to-end STREAM packet data (i.e., packets that should only be visible between
+   *                                sender and receiver).
+   * @param link                    A {@link Link} that is used to send ILPv4 packets to an immediate peer.
+   * @param executorService         executorService to run the payments
+   * @param sendPacketSleep amount of time for a thread to sleep before sending more packets
+   */
+  public SimpleStreamSender(
+    final StreamEncryptionService streamEncryptionService,
+    final Link link,
+    ExecutorService executorService,
```
sappenin: Minor nit:
```suggestion
final ExecutorService executorService,
```

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -120,33 +146,52 @@ public SimpleStreamSender(final StreamEncryptionService streamEncryptionService,
...
   /**
    * Required-args Constructor.
    *
-   * @param streamEncryptionService A {@link StreamEncryptionService} used to encrypt and decrypted end-to-end STREAM
+   * @param streamEncryptionService An instance of {@link StreamEncryptionService} used to encrypt and decrypted
+   *                                end-to-end STREAM packet data (i.e., packets that should only be visible between
+   *                                sender and receiver).
+   * @param link                    A {@link Link} that is used to send ILPv4 packets to an immediate peer.
+   * @param executorService         executorService to run the payments
+   * @param sendPacketSleep amount of time for a thread to sleep before sending more packets
```
sappenin: Minor nit: formatting

```suggestion
   * @param         sendPacketSleep amount of time for a thread to sleep before sending more packets
```

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -120,33 +146,52 @@ public SimpleStreamSender(final StreamEncryptionService streamEncryptionService,
...
+  /**
+   * Required-args Constructor.
+   *  @param streamEncryptionService A {@link StreamEncryptionService} used to encrypt and decrypted end-to-end STREAM
    *                                packet data (i.e., packets that should only be visible between sender and
    *                                receiver).
    * @param link                    A {@link Link} that is used to send ILPv4 packets to an immediate peer.
    * @param executorService         A {@link ExecutorService} to run the payments.
    * @param streamConnectionManager A {@link StreamConnectionManager} that manages connections for all senders and
-   *                                receivers in this JVM.
+   * @param sendPacketSleep amount of time for a thread to sleep before sending more packets
```
sappenin:
```suggestion
   * @param                         sendPacketSleep amount of time for a thread to sleep before sending more packets
```

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -120,33 +146,52 @@ public SimpleStreamSender(final StreamEncryptionService streamEncryptionService,
...
-      final StreamEncryptionService streamEncryptionService,
-      final Link link,
-      final ExecutorService executorService,
-      final StreamConnectionManager streamConnectionManager
-  ) {
+    final StreamEncryptionService streamEncryptionService,
+    final Link link,
+    final ExecutorService executorService,
+    final StreamConnectionManager streamConnectionManager,
+    Optional<UnsignedLong> sendPacketSleep) {
```
sappenin:
```suggestion
  final Optional<UnsignedLong> sendPacketSleep
  ) {
```

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -298,6 +347,9 @@ default int totalPackets() {
       this.receiverDenomination = Optional.empty();
 
       this.unrecoverableErrorEncountered = new AtomicBoolean(false);
+
+      Objects.requireNonNull(sendPacketSleep);
+      this.sendPacketSleep = sendPacketSleep.orElse(UnsignedLong.valueOf(100L)).longValue();
```
sappenin: There was talk that we should make this default much lower, like 10 -- @nhartner @theotherian does that sound OK?

stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderTests.java
```diff
@@ -65,7 +66,13 @@ public void constructThreeArgWithNullLink() {
   @Test
   public void constructThreeArgWithNullExecutor() {
     expectedException.expect(NullPointerException.class);
-    new SimpleStreamSender(streamEncryptionServiceMock, linkMock, null);
+    new SimpleStreamSender(streamEncryptionServiceMock, linkMock, (ExecutorService) null);
+  }
+
+  @Test
+  public void constructThreeArgWithNullSleep() {
```
sappenin: Not sure if it's worth it, but should we make a unit test with a simulated Clock to test the loop?

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
nkramer44: I set the default to 10ms.  After running some tests, there doesn't seem to be a difference in total send time between `SimpleStreamSender`s with a 10ms sleep and 5ms sleep.

We should maybe do some more testing to figure out the most optimal sleep time, but that's probably overkill for now

stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderTests.java
nkramer44: Check out the latest commit to `SimpleStreamSenderIT`.  I send the same payment through a `SimpleStreamSender` with sleep time set to 100ms and a `SimpleStreamSender` with sleep time set to 5ms.  Total time to send the payments (in this case 7 packets) was much much lower with a smaller sleep time

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
sappenin: Per the Quilt community call - let's explore 20ms or 50ms and just double-check that 10ms seems to be the best sleep time.

stream-parent/stream-client/src/test/java/org/interledger/stream/sender/SimpleStreamSenderTests.java
sappenin: Per the ILP community call, consensus is that this is probably not worth the effort.

stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
nkramer44: Ran a bunch of different sleep times.  10ms always seems to execute faster than anything > 10ms.  Once you get below 10ms, it's a wash which one will be faster.

Note this was tested on a payment that sends roughly 20 packets.

PR #436, opened 2020-02-25
support multilateral mode from JS ilp-http-plugin

nhartner:


link-parent/link-ilp-over-http/src/test/java/org/interledger/link/http/MultilateralUrlFactoryTest.java
```diff
@@ -0,0 +1,43 @@
...
+    MultilateralUrlFactory factory = new MultilateralUrlFactory(() -> OPERATOR, NON_MULTILATERAL_URL);
+    assertThat(factory.getOutgoingUrl(OPERATOR.with("spsp").with("0").with("secretstuff")))
+        .isEqualTo(NON_MULTILATERAL_URL);
+    assertThat(factory.getOutgoingUrl(OPERATOR.with("spsp").with("ONE").with("secretstuff")))
+        .isEqualTo(NON_MULTILATERAL_URL);
+  }
+
+  @Test
+  public void getOutgoingUrlWithNonMultilateralAddress() {
+    MultilateralUrlFactory factory = new MultilateralUrlFactory(() -> OPERATOR, MULTILATERAL_URL);
```
sappenin: Are we sure this is how we want this to function? Won't the Link blow up when it tries to resolve `https://www.test.com/spsp-%/ilp`?

sappenin: @nhartner and I discussed and we decided that this will end up in a REJECT anyway. If an infra operator wanted to handle URLs with a percent-sign like this, they could. 

If this becomes a pain-point, we could specify a `defaultURL` here that doesn't include a `%` wildcard so that if a destination address cannot be mapped, then at least the HTTP call might succeed. However, not going to do this now because it's ambiguous if anyone will need it.

nhartner:
deferred for now since multilateral support requires non-trivial refactoring in the connector to support ildcp with multilateral

PR #437, opened 2020-03-09
Make `WELL_KNOWN` constant public

keefertaylor:
I'd like to make WELL_KNOWN a public constant. This is useful because it means that there is a source of truth for this value. 

In practice, this improves our test code because we no longer have to hard code this magic constant.


keefertaylor:
AFAICT, the Java14 build is broken irrespective of this change.

sappenin:
> AFAICT, the Java14 build is broken irrespective of this change.

I fixed this in #438, so once we merge that, this issue should resolve for you.

PR #440, opened 2020-03-17
Stop sending packets if F-family error encountered

### LINKED ISSUE #430: STREAM Sender: Unreachable destinations will retry until timeout but should short-circuit
If a STREAM sender tries to send a payment on a link but the Connector can't forward the packets (i.e., the StreamSender gets an `F02 Unreachable` error, the Sender should stop sending STREAM packets. However, the current implementation retries every 10ms until the overall sendMoney timeout is hit.

nkramer44:
Fixes #430

### DIFF
--- a/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
+++ b/stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
@@ -1,6 +1,7 @@
 package org.interledger.stream.sender;
 
 import static org.interledger.core.InterledgerErrorCode.F00_BAD_REQUEST;
+import static org.interledger.core.InterledgerErrorCode.F08_AMOUNT_TOO_LARGE;
 import static org.interledger.core.InterledgerErrorCode.F08_AMOUNT_TOO_LARGE_CODE;
 import static org.interledger.core.InterledgerErrorCode.T04_INSUFFICIENT_LIQUIDITY_CODE;
 import static org.interledger.stream.StreamUtils.generatedFulfillableFulfillment;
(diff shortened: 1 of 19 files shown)


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -482,17 +482,26 @@ default int totalPackets() {
...
+                                                                     StreamPacket streamPacket) {
       InterledgerResponsePacket response = link.sendPacket(preparePacket);
-      response.handle((fulfill) -> {}, (reject) -> {
-        if (reject.getCode().equals(F00_BAD_REQUEST)) {
-          unrecoverableErrorEncountered.set(true);
-        }
+      response.handle(
+        (fulfill) -> {},
+        (reject) -> {
+          if (!reject.getCode().equals(F08_AMOUNT_TOO_LARGE) &&
```
kincaidoneil: Shouldn't this also keep retrying on F99 errors?

nkramer44: @sappenin would know better...  

The Java receiver only returns an F99 if the prepare amount < stream amount, but the Java sender doesn't have any retry behavior to deal with F99s.  Maybe it should?

kincaidoneil: There are other cases where the JavaScript STREAM receiver returns F99 errors, too, which is something to consider.

> The Java receiver only returns an F99 if the prepare amount < stream amount

But should the payment fail if only a single packet doesn't meet the minimum exchange rate? Or should there be multiple attempts?

Rust and JS retry like normal on F99s, but I think we probably need more sophisticated behavior for the exchange rate issues.

nkramer44: Agreed. For now, our main motivation for short circuiting is to reduce our log output/error alerting, which is mainly happening on F02s. I will narrow this condition to only short circuit on F02s, and we can discuss what we want to do on F99s.

sappenin:
This PR has been superseded by https://github.com/hyperledger/quilt/pull/443

PR #443, opened 2020-03-30
Fixes #430

### LINKED ISSUE #430: STREAM Sender: Unreachable destinations will retry until timeout but should short-circuit

sappenin:
* Supersedes PR #440 (now closed).
* Fixes #430: Clarify and enhance the InterledgerErrorCodes that the SimpleStreamSender will fail-fast on and stop the stream. These include any F or R-family error, except F08 and F99. Any T-family error will NOT abort the sendMoney operation.
* Fixes Invalid Denomination: If a receiver doesn't send back `ConnectionAssetDetails` frames to the sender during preflight, then the denomination will be absent. In this case, the FixedReceiverAmountPaymentTracker doesn't handle this condition properly.
* Enhance unit & integration tests to cover these scenarios.
* Add Javadoc in various places and cleanup formatting


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -804,28 +835,90 @@ void handleReject(
...
+     *                     response to a prepare packet.
+     */
+    @VisibleForTesting
+    void checkForAndTriggerUnrecoverableError(final InterledgerRejectPacket rejectPacket) {
+      Objects.requireNonNull(rejectPacket);
+
+      if (ErrorFamily.FINAL.equals(rejectPacket.getCode().getErrorFamily())) {
+        // Most Final errors trigger immediate stoppage of send operations.
+        if (!rejectPacket.getCode().equals(F08_AMOUNT_TOO_LARGE) &&
+            !rejectPacket.getCode().equals(F99_APPLICATION_ERROR)) {
```
nhartner: Seeing these 2 error codes repeated. Might make sense to create `static final List<InterledgerErrorCode> retryableFinalErrors = Lists.newArrayList(F08, F99)` and then check if `retryableFinalErrors.contains(rejectPacket.getCode())`

sappenin: This is fixed.

PR #447, opened 2020-04-10
Support standard NIST-recommended AuthTag ByteOrdering in STREAM Encryption Service

### LINKED ISSUE #237: Change AuthTag ByteOrdering in STREAM RFC
File issue in Rust and JS and Quilt to not [re-arrange the bytes](https://github.com/hyperledger/quilt/blob/master/stream-parent/stream-core/src/main/java/org/interledger/stream/crypto/JavaxStreamEncryptionService.java#L72) - the tag should go last so that we can easily conform to (standard?) or at least typical library implementations.

sappenin:
Fixes #237

* Fix the AuthTag byte ordering in the default StreamEncryptionService in a fully-backward compatible manner.
* Deprecate JavaxStreamEncryptionService and rename to the more appropriate AesGcmStreamEncryptionService.
* Add unit tests.


stream-parent/stream-core/src/main/java/org/interledger/stream/crypto/AesGcmStreamEncryptionService.java
```diff
@@ -0,0 +1,375 @@
...
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+/**
+ * An {@link StreamEncryptionService} that uses a JavaKeystore for underlying key storage.
+ */
+public class AesGcmStreamEncryptionService implements StreamEncryptionService {
```
sappenin, line 30: This class is a copy/paste of the original JavaxStreamEncryptionService, with added support for encryption/decryption modes to support existing functionality in a fully backwards-compatible manner, plus standards-compliant functionality.

PR #457, opened 2020-04-26
Use cached thread-pool instead of unbounded pool

sappenin:
Using Executors.newFixedThreadPool will start off with 30 threads, but will create new threads in an unbounded fashion. Instead, using the cachedThreadPool this behavior will find a steady-state.


nhartner:
https://www.baeldung.com/java-executors-cached-fixed-threadpool

> The cached pool starts with zero threads and can potentially grow to have Integer.MAX_VALUE threads.

sappenin:
Closing this PR due to me not thinking.

Issue #308, opened 2019-09-30
Revisit StreamSender ConnectionClosing


sappenin:
The implementation of `SimpleStreamSender` doesn't ever close the STREAM Connection nor the Stream itself. 

This is because the current `ConnectionManager`, while it _can_ track sequences properly for parallel `sendMoney` calls, does not control the _actual_ opening/closing of a Connection using frames (instead, all frames are currently sent by the SimpleStreamSender).

Thus, multiple streams in parallel can cause problems because while the sequence number of a Connection will be consistent, the SimpleStreamSender only really supports a single SendMoney StreamId (i.e., 1).

Thus, we should enhance the `SimpleStreamSender` to do the following:

* Make `sendMoneyPacketized` close the  _STREAM_ after a payment is completed (but not the Connection).
* Return a `StreamClosedResult` from `sendMoneyPacketized`.
* Consider when and how a StreamConnection should be closed. We could introduce a new method into the StreamSender interface called `closeConnection()` and expect outside callers to close this. Alternatively, we could introduce a smarter ConnectionManager that can detect when a connection should be closed.

Whatever we decide, we should be sure that  we don't remove that Connection when this happens -- in other words, once a connection is closed, it should not be re-used in order to protect the AES-GCM requirements of IL-RFC-29.

sappenin:
SimpleStreamSender will soon be deprecated in-favor of [ILP Pay](https://github.com/hyperledger/quilt/tree/df/ilp-pay) functionality. Closing this.

PR #464, opened 2020-05-14
send stream close frame when all is said and done on a send money request

nhartner:


stream-parent/stream-client/src/main/java/org/interledger/stream/sender/SimpleStreamSender.java
```diff
@@ -964,100 +971,94 @@ void setUnrecoverableErrorEncountered(boolean unrecoverableErrorEncountered) {
...
+       final InterledgerPreparePacket preparePacket = InterledgerPreparePacket.builder()
+           .destination(destinationAddress)
+           .amount(UnsignedLong.ZERO)
+           .executionCondition(executionCondition)
+           .expiresAt(DateUtils.now().plusSeconds(30L))
+           .data(encryptedStreamPacket)
+           .typedData(streamPacket)
+           .build();
+
+       link.sendPacket(preparePacket);
```
sappenin, line 1039: Should this not be done in  a handler, like this?

```
link.sendPacket(preparePacket).handle(
                  fulfillPacket -> ...,
                  rejectPacket -> ...
```

Put another way, if the `link.sendPacket` fails, should lines 1042-1061 still occur? If the answer is "yes" then those should be moved above line 1039. If the answer is "no", then this seems fine, but feels more natural to only close the stream if the other side replies that it actually got those frames.

nhartner: Whether they fulfill or reject, our sender is done using the stream connection. It's more like an FYI to the receiver at this point. I didn't see the point in adding any logic there. But maybe I don't understand this well enough to see the reason that's incorrect.

sappenin: Ok, in that case lines 1041-1061 should come before `link.sendPacket` on line 1039. Rationale is that if line 1039 fails, then the `streamConnection` won't actually be marked as closed.

nhartner: I just remove or comment out lines 1041-1061 (again) because my changes only send a StreamClose frame but not the ConnectionClose frame.  I only uncommented the code to send a StreamClose frame. The code to send ConnectionClose frame is still commented out because that felt like it had bigger issues according the the github issue #308. I don't understand the implications, but I would've expected that I shouldn't use the stream connection after it's been closed. And moving lines 1041-1046 before sending the packet would do just that.

nhartner: commenting out for now

PR #478, opened 2021-01-24
Introduce ILP-Pay

sappenin:
ILP Pay is a port of the [JS variant with the same name](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay). It is meant to replace the SimpleStreamSender because it both improves overall Stream-sending logic, but also more accurately support path-based exchange rate calculations.


sappenin:
Closing this in-favor of https://github.com/hyperledger/quilt/pull/479/

